PR #1, opened 2022-08-29
DM-31226: Add WCSFit task

cmsaunders:


parejkoj (commented):
I think we'll want a separate testdata package for this: I see a bunch of FITS files in `tests/data/`, and it looks like those take up ~5MB right now.

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+        dtype=float,
+        default=1.0
+    )
+    minMatches = pexConfig.Field(
+        doc="Number of matches required to keep a source object",
+        dtype=int,
+        default=2
+    )
+    allowSelfMatches = pexConfig.Field(
+        doc="Allow multiple object matches from the same viist",
```
erykoff, line 271: `viist` -> `visit`.  Also, I don't understand this doc string.

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+        storageClass="DataFrame",
+        dimensions=("instrument", "skymap", "tract"),
+    )
+
+
+class WCSFitConfig(pipeBase.PipelineTaskConfig,
+                   pipelineConnections=WCSFitConnections):
+    """Configuration for WCSFitTask"""
+    sourceSelector = sourceSelectorRegistry.makeField(
+        doc="How to select sources for cross-matching",
```
erykoff, line 253: doc strings in configs should have a trailing period.  Also above for connections.

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+        default=False
+    )
+    sourceFluxType = pexConfig.Field(
+        dtype=str,
+        doc="Source flux field to use in source selection and to get fluxes from the catalog.",
+        default='apFlux_12_0'
+    )
+    systematicError = pexConfig.Field(
+        dtype=float,
+        doc="Systematic error padding for the science images (marcsec)",
```
erykoff, line 282: Is this a padding, or something added in quadrature?

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+        dtype=int,
+        doc="Order of device polynomial model",
+        default=4
+    )
+    exposurePolyOrder = pexConfig.Field(
+        dtype=int,
+        doc="Order of exposure polynomial model",
+        default=6
+    )
+    useProperMotion = pexConfig.Field(
```
erykoff, line 315: I think this should be `doFitProperMotion` not `useProperMotion` which I assume is "used" when `requireProperMotion` is set below?

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+        wcss.append(refWcs)
+
+        extensionVisits.append(len(exposureNames) - 1)
+        extensionDetectors.append(-1)  # REFERENCE device must be -1
+        visits.append(-1)
+        detectors.append(-1)
+        extensionType.append('REFERENCE')
+
+        # Make a table of information to use elsewhere in the class
+        self.extensionInfo = pd.DataFrame({'visit': visits, 'detector': detectors,
```
erykoff, line 632: Does this need to be a dataframe, or could it be a dict of numpy arrays/lists (in the case of the wcss).

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+
+        selected = self.referenceSelector.run(skyCircle.refCat)
+        # Need memory contiguity to get reference filters as a vector.
+        if not selected.sourceCat.isContiguous():
+            refCat = selected.sourceCat.copy(deep=True)
+        else:
+            refCat = selected.sourceCat
+
+        if self.config.useProperMotion:
+            # GBDES cannot currently accommodate catalogs with a mixture of PM
```
erykoff, line 678: Is there a ticket for this?  Or is this okay because all the gaia stars have a PM measurement?  Also, is this a limitation on all the stars, or just the refstars?

tests/test_wcsfit.py
```diff
@@ -0,0 +1,547 @@
...
+        # Set up the schema for the output catalogs
+        schema = afwTable.ExposureTable.makeMinimalSchema()
+        schema.addField('visit', type='L', doc='Visit number')
+
+        catalog = afwTable.ExposureCatalog(schema)
+        catalog.resize(len(self.detectors))
+        catalog['visit'] = visit
+
+
+class MockInputCatalogRef():
```
erykoff, line 55: There is a better way of doing this, with an `lsst.pipe.base.InMemoryDatasetHandle`.

ups/drp_tasks.table
```diff
@@ -3,6 +3,7 @@
 # - Common third-party packages can be assumed to be recursively included by
 #   the "base" package.
 setupRequired(base)
+#setupRequired(gbdes)
```
erykoff, line 6: I think this shouldn't have the leading `#`?

erykoff (commented):
This is a first batch of comments on the code; I will have more.

python/lsst/drp/tasks/wcsfit.py
cmsaunders: This is a `gbdes` config option. I changed this to "Allow multiple sources from the same visit to be associated with the same object". If this if false, then the association step throws out any objects that have multiple sources from the same exposure.

python/lsst/drp/tasks/wcsfit.py
cmsaunders: It's added in quadrature, because it gets converted to covariance. Do you think it should be called something else?

python/lsst/drp/tasks/wcsfit.py
cmsaunders: This makes me unhappy because the doc strings are not sentences. Do you know where it is in the dev guide so I can try to understand this rule?

python/lsst/drp/tasks/wcsfit.py
cmsaunders: `requireProperMotion` sets a config parameter for the reference object loader, but I had a misunderstanding about what it did, which is just to check that the proper motion parameters are in the schema, not that they are actually there. I've changed this so that 'useProperMotion' just also sets `requireProperMotion`. But I agree that the name should be changed.

cmsaunders: I changed it to `fitProperMotion`. Let me know if you feel strongly about `doFitProperMotion`.

python/lsst/drp/tasks/wcsfit.py
cmsaunders: Yeah, I guess I didn't end up using any dataframe-specific functionality.

python/lsst/drp/tasks/wcsfit.py
cmsaunders: I've actually been experimenting with this, and I think that `gbdes` can handle a mix of pm and non-pm sources. I asked Gary about it, but haven't heard back yet. I think the way Gary was running he only used Gaia sources with proper motion, but you do lose a lot (about 40% in the footprint of the test data I've been using). I'll make a ticket depending on what he says.

tests/test_wcsfit.py
cmsaunders: `lsst.pipe.base.InMemoryDatasetHandle` doesn't work because I need a `.ref` property. I played around with different methods of making something that would work in `lsst.meas.algorithms.ReferenceObjectLoader`, but I gave up and did this. I guess adding `.ref` to `InMemoryDatasetHandle` would be another option.

erykoff: Upon some follow-up, this should _not_ have a public `ref` property at all.  I'm curious what was throwing an error, or is it something just in the testing here?

erykoff: But I realize this is a bit outside of this ticket.  However, I think that we have some not great test code that has just been cargo culted around.  It might need a follow-up ticket.

timj: If a `PipelineTask.run()` method somewhere needs to know the `DatasetRef` of the dataset then that seems very wrong. I can't think of a reason a `run()` method would ever need to know that. It's public because it's a dataclass and so needs to be passed in to the constructor. Do we know what code is trying to look at the ref and why it's wanting to know?

parejkoj: Why do you have all this infrastructure for making fake catalogs and datasetRefs, instead of using the testdata_jointcal package that has butler datasets ready to go?

@timj : not a topic for discussion on this ticket, but `ReferenceObjectLoader` needs the refs in its `__init__`.

erykoff: @parejkoj I think that the idea is with sim data here there are additional tests that are possible when you know the truth rather than for the real data in testdata_jointcal.

Also, the `ReferenceObjectLoader` needs the deferred dataset handles, not the `ref`.  I just did a quick modification locally to `test_wcsfit.py` and (a) added `from lsst.pipe.base import InMemoryDatasetHandle`; (b) deleted `MockInputCatalogRef`; (c) changed `inputCatalogRef = MockInputCatalogRef(visitSourceTable, dataId={"visit": visit})` to `inputCatalogRef = InMemoryDatasetHandle(visitSourceTable, storageClass="DataFrame", dataId={"visit": visit})`; (d) changed `deferredRefCat = MockInputCatalogRef(refCat, ref=refDataId.ref)` to `deferredRefCat = InMemoryDatasetHandle(refCat, storageClass="SourceCatalog")` and the tests ran fine.

cmsaunders: @parejkoj and @erykoff, I had to remind myself why I'm not using `testdata_jointcal`. `testdata_jointcal` would work in place of the included test data. Realistic WCSs and observation parameters are all that are needed; the source tables are simulated. However, I was specifically asked not to use `testdata_jointcal` because it has `obs_subaru` and other obs packages as dependencies, and we didn't want to tie `drp_tasks` to all of those packages.

parejkoj: You would make it an optional dependency of this package, not a required one, and then have tests that check for its existence (as is done in fgcmcal and jointcal).

If you're going to not use testdata_jointcal (but I really think you should, since we've characterized that data pretty well), you need to move the test files out of this package and into an optional test data package anyway; we can't have large FITS files in code packages, as you're doing here.

ups/drp_tasks.table
parejkoj, line 6: You clearly have more dependencies than are listed here: I count at least 6 in `wcsfit.py`.

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+                dataId=visit_data_id,
+                findFirst=True,
+            ).expanded()
+        )
+    yield from refs
+
+
+def makeRefCovarianceMatrix(refCat, inputUnit=u.radian, outputCoordUnit=u.degree, outputPMUnit=u.marcsec):
+    """Make a covariance matrix for the reference catalog including proper
+    motion and parallax.
```
parejkoj, line 84: I think this docstring needs some more explanation, because it looks like you're making a very specific kind of flattened covariance matrix. I assume this is specifically formatted for gbdes?

You also don't need `inputUnit`, since the refcat columns already have units attached.

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+        Degree of the polynomial.
+
+    Returns
+    -------
+    astPoly : `astshim.PolyMap`
+        Coefficients in AST polynomial format
+    """
+    polyArray = np.zeros((len(coefficients), 4))
+    N = len(coefficients) / 2
+    degree = int(-1.5 + 0.5 * (1 + 8 * N)**0.5)
```
parejkoj, line 159: What is this math? I think this definitely needs a comment explaining what is going on here; that int conversion of an inherently floating point number (sqrt of an odd number?) worries me.

python/lsst/drp/tasks/wcsfit.py
parejkoj: Field docstrings should be sentences (potentially multiple); they are built into the `Configuration Fields` section of the pipeline docs. Our examples in `pex.Config` are inconsistent on punctuation and capitalization: I'll file a ticket to get that cleaned up.

parejkoj: https://jira.lsstcorp.org/browse/DM-36696

python/lsst/drp/tasks/wcsfit.py
cmsaunders: This is the quadratic formula applied to the formula for finding the number of coefficients of a multivariable polynomial up to a certain degree (I forget what the latter formula is called). It's guaranteed to be an integer. I added a comment.

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+    ConfigClass = WCSFitConfig
+    _DefaultName = "wcsfit"
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.makeSubtask("sourceSelector")
+        self.makeSubtask("referenceSelector")
+
+        self.detectors = []
+        self.visits = []
```
parejkoj, line 374: It is strongly recommended that Tasks be stateless. You hold other instance variables later in this code as well (e.g. `extensionInfo`); can you restructure it to get rid of that state? See the final paragraph of this section:

https://pipelines.lsst.io/v/daily/modules/lsst.pipe.base/creating-a-task.html#the-run-method

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+                componentDict = {"Type": "Poly",
+                                 "XPoly": {"OrderX": self.config.devicePolyOrder,
+                                           "SumOrder": True},
+                                 "YPoly": {"OrderX": self.config.devicePolyOrder,
+                                           "SumOrder": True},
+                                 "XMin": xMin, "XMax": xMax, "YMin": yMin, "YMax": yMax}
+            elif "identity" in component.lower():
+                componentDict = {"Type": "Identity"}
+            else:
+                raise ValueError(f"deviceModel component {component} is not supported")
```
parejkoj, line 812: These are configs, so I think you should validate them as part of `config.validate()`, so that the failure happens on task init.

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+        wcsf : `wcsfit.WCSFit`
+            WCS-fitting object
+        inputCatalogRefs : `list`
+            List of DeferredDatasetHandles pointing to visit-level source
+            tables
+        """
+        for inputCatalogRef in inputCatalogRefs:
+            visit = inputCatalogRef.dataId['visit']
+
+            inputCatalog = inputCatalogRef.get(parameters={'columns': self.columns})
```
parejkoj, line 849: It looks like you're reading these catalogs twice, once here and once in `_load_catalogs_and_associate`. Is that an intentional tradeoff between memory and disk reads? I'd be rather surprised that the catalogs themselves are a significant part of the memory footprint: for jointcal, the full fit matrix was vastly larger, and the catalogs basically were irrelevant in comparison.

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+
+                extensionIndex = self.extensionInfo[(self.extensionInfo['visit'] == visit)
+                                                    & (self.extensionInfo['detector'] == detector)].index[0]
+                sourceCat = detectorSources[self.sourceIndices[extensionIndex]]
+
+                d = {"x": sourceCat['x'].to_numpy(),
+                     "y": sourceCat['y'].to_numpy(),
+                     "xCov": sourceCat['xErr'].to_numpy()**2,
+                     "yCov": sourceCat['yErr'].to_numpy()**2,
+                     "xyCov": np.zeros(len(sourceCat))}
```
parejkoj, line 863: Jointcal computed the xy err from the x and y errors and the shape moments (see `CcdImage:loadCatalog`): would a similar assumption be better here, than just setting them to zero?

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+        iwcToSkyWcs = afwgeom.makeSkyWcs(lsst.geom.Point2D(0, 0), tangentPoint, cdMatrix)
+        iwcToSkyMap = iwcToSkyWcs.getFrameDict().getMapping("PIXELS", "SKY")
+        skyFrame = iwcToSkyWcs.getFrameDict().getFrame("SKY")
+
+        frameDict = astshim.FrameDict(pixelFrame)
+        frameDict.addFrame("PIXELS", normMap, normedPixelFrame)
+
+        currentFrameName = "NORMEDPIXELS"
+
+        # Assume python 3.7+ so dict is ordered:
```
parejkoj, line 934: We're on py3.10 now, so this comment isn't necessary.

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+        Parameters
+        ----------
+        mapDict : `dict`
+            Dictionary of mapping parameters
+        centerRA : `lsst.geom.Angle`
+            RA of the tangent point
+        centerDec : `lsst.geom.Angle`
+            Declination of the tangent point
+        doNormalizePixels : `bool`
+            Whether to normalize pixels so that range is [-1,1]
```
parejkoj, line 897: I just noticed this, but all of your method docstrings should be sentences, with punctuation.

https://developer.lsst.io/python/numpydoc.html#parameters

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+            if m == len(mapDict) - 1:
+                newFrameName = "IWC"
+            else:
+                newFrameName = "INTERMEDIATE" + str(m)
+            newFrame = astshim.Frame(2, f"Domain={newFrameName}")
+            frameDict.addFrame(currentFrameName, astMap, newFrame)
+            currentFrameName = newFrameName
+        frameDict.addFrame("IWC", iwcToSkyMap, skyFrame)
+
+        outWCS = afwgeom.SkyWcs(frameDict)
```
parejkoj, line 955: If you're going to use camelCase throughout the code, this should be `outWcs`. I would suggest switching to snake_case throughout, though (see [RFC-623](https://jira.lsstcorp.org/browse/RFC-623)).

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+            Factor by which to normalize x-dimension. Corresponds to width of
+            detector
+        yScale : `float`
+            Factor by which to normalize y-dimension. Corresponds to height of
+            detector
+
+        Returns
+        -------
+        outWCS : `lsst.afw.geom.SkyWcs`
+            WCS constructed from the input mappings
```
parejkoj, line 908: Looking at the underlying code, the WCS this returns does not have a FOCAL frame, as our camera convention assumes. Is that what `INTERMEDIATE` is supposed to be? If there isn't a FOCAL frame, we may have difficulty integrating these output WCSs with the rest of our tooling (e.g. for computing radial distortions).

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+        wcsf : `wcsfit.WCSFit`
+            WCSFit object, assumed to have fit model
+        inputVisitSummary : `lsst.afw.table.ExposureCatalog`
+            Catalogs with per-detector summary information from which to grab
+            detector information
+
+        Returns
+        -------
+        catalogs : `dict`
+            Dictionary of `lsst.afw.table.ExposureCatalog` objects with the WCS
```
parejkoj, line 972: What are the keys and values of this dict?

https://developer.lsst.io/python/numpydoc.html#dictionary-types

parejkoj (dismissed):
Since I started commenting on other things, I looked through wcsfit.py and have several requested changes.

You use a mix of camelCase and snake_case here: can you please choose one (we're moving new code to snake_case, post-[RFC-623](https://jira.lsstcorp.org/browse/RFC-623)) and use it consistently? Your method names have both (and also mixes, like `_load_refCat`!).

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+        default=6
+    )
+    useProperMotion = pexConfig.Field(
+        dtype=bool,
+        doc="Fit the proper motions of the objects",
+        default=False
+    )
+    verbose = pexConfig.Field(
+        dtype=int,
+        doc="Verbosity level=0, 1, or 2",
```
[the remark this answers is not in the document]
timj: If gbdes is using `print()` and not `logging` then there are going to be all sorts of wrinkles with capturing the output in batch (we capture logs but where stdout goes depends on the workflow system).

Is it possible to map these verbosity levels to WARN, INFO, VERBOSE log levels and so derive the verbosity level based on the logging level that is active for your task code? Ideally gbdes would use logging properly (although you might still have to work out how to set the default log level for it since we don't set non-lsst logger levels by default.

python/lsst/drp/tasks/wcsfit.py
erykoff: First of all, I agree with using snake_case for new code.  However, the LSST dev guide is WRONG on this front re: camelCase, see PEP-8: https://peps.python.org/pep-0008/ "Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError."

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+        result : `lsst.pipe.base.Struct`
+            ``outputWCSs`` : `list` of `lsst.afw.table.ExposureCatalog`
+                List of exposure catalogs (one per visit) with the WCS for each
+                detector set by the new fitted WCS
+            ``fitModel`` : `wcsfit.WCSFit`
+                Model-fitting object with final model parameters
+            ``outputCatalog`` : `pd.DataFrame`
+                Catalog with fit residuals of all sources used
+        """
+        self.log.info("Gathering instrument, exposure, and field info")
```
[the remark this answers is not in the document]
timj: If it helps, in batch we run with "verbose" log level enabled and not debug level. By default a user will see only info log messages from their own pipetask run.

python/lsst/drp/tasks/wcsfit.py
erykoff: That is an interesting point.  And I would say that this should not be True for HSC, the epoch from field to field varies by up to several years, and this does show up in the Gaia residuals if you're not careful.   (I've been ascribing this to PM; maybe wrong...?). But the HSC observations are not continuous, and the mean epoch of any field can be quite different from Gaia reference epoch, so the same argument holds.

tests/test_wcsfit.py
cmsaunders: The `wcsfit.Instrument` is not related to the `obs.base.Instrument`. It is just a `gbdes` class that holds basic information, like the number of detectors and their dimensions (in pixels). The detector layout comes from the input WCS.

parejkoj: Where does the number of detectors and their dimensions come from then? You should be getting that from the  Camera object.

python/lsst/drp/tasks/wcsfit.py
cmsaunders: The covariance matrix is just a np.ravel(order='C')-style flattening. I added to the documentation.

I've had some trouble before pulling the units out of the `simpleCatalog`, I finally figured out I could do
```
unit = refCat.schema[<key>].asField().getUnits()
```
However, this only works for the keys where the units are listed when you print the schema. For 'coord_ra', for example, when you look at the table you see 'rad' under 'coord_ra' in the column name, so obviously the information is there, but I can't figure out how to access this.

erykoff: You just have to "know" that the `coord_ra` and `coord_dec` are in radians if you are using the array access rather than the unbelievably slow row-by-row access.  But note that that statement only applies to afw tables (like the refcats); for parquet tables sometimes they're in degrees.

python/lsst/drp/tasks/wcsfit.py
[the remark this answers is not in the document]
cmsaunders: All the output from `gbdes` is from c++ and I think it's all from `std::cerr`. It doesn't currently use any logger, it just prints to standard error depending on the verbosity level (which could be matched to WARN, INFO, etc).  @timj, do you know if there is any way to capture this properly without using `lsst.log`?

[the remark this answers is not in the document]
timj: Standard error is going to be caught by the pipeline infrastructure and will end up somewhere but it won't be stored in the butler logs or available to the real-time logging system. I think the C++ stderr is not the same as the python stderr so capturing it and forwarding it to a logger is a bit tricky (I think @ktlim tried to get something working for that when we used to use lsst.log in the past in jupyter notebooks). The best way is to start using the C++ lsst.log package in gbdes.

python/lsst/drp/tasks/wcsfit.py
cmsaunders: Ok, I'm going to set it to exclude non-PM objects for now, though I think it would be worth doing a small study in the future to see what the bias vs. improvement in fit tradeoff is for HSC.

python/lsst/drp/tasks/wcsfit.py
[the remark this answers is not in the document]
cmsaunders: I removed the config parameter and now set `verbose` for `gbdes` from `self.log.getEffectiveLevel()`. I'm going to make a jira ticket for improving the output in `gbdes`.

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+        # Use only stars because aperture fluxes of galaxies are biased and
+        # depend on seeing.
+        self.sourceSelector["science"].doUnresolved = True
+        self.sourceSelector["science"].unresolved.name = "extendedness"
+
+        # Use only isolated sources.
+        self.sourceSelector["science"].doIsolated = True
+        self.sourceSelector["science"].isolated.parentName = "parentSourceId"
+        self.sourceSelector["science"].isolated.nChildName = "deblend_nChild"
+        # Do not trust either flux or centroid measurements with flags,
```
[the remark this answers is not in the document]
cmsaunders: Yes, I modified the language, but note that this is copied verbatim from {{jointcal}}.

python/lsst/drp/tasks/wcsfit.py
cmsaunders: I tried implementing this, following the `loadCatalog` method, but I ran into some problems. There are a fair number (~1% maybe) of detections that have inferred covariances that are too high compared to the variance. I see this throws a warning in `jointcal`, and if I leave these objects in, the final chi2 is catastrophically worse. I could remove these from the fit; there may be a flag that helps. I don't know if the bad shape measurement is a symptom of other problems. @parejkoj, did you ever look into this?

python/lsst/drp/tasks/wcsfit.py
cmsaunders: I rewrote this to explain why the order matters.

python/lsst/drp/tasks/wcsfit.py
cmsaunders: Yeah, I don't want to keep the catalogs in memory, and the disk read time is relatively short (about 0.05 seconds per visit). I haven't tested how it affects performance, but it definitely would be a significant part of the memory footprint. Remember that a big difference between `jointcal` and `gbdes` is that the star positions are not fit parameters in `gbdes`, so the fit matrix is actually not that big.

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1019 @@
...
+        # Add the science and reference sources
+        self._addObjects(wcsf, inputCatalogRefs)
+        self._addRefObjects(wcsf)
+
+        # Do the WCS fit
+        wcsf.fit()
+        self.log.info("WCS fitting done")
+
+        outputWCSs = self._makeOutputs(wcsf, inputVisitSummary[0])
+        outputCatalog = pd.DataFrame(wcsf.getOutputCatalog())
```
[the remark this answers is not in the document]
cmsaunders: This mimics the input catalogs ("preSourceTable_visit", which are also DataFrames. This output exists to facilitate doing diagnostics, and is otherwise not used downstream.

python/lsst/drp/tasks/wcsfit.py
cmsaunders: Ok, I'm not changing all the internal variables, but I made the API all snake_case.

python/lsst/drp/tasks/wcsfit.py
parejkoj: This is related to DM-7101 and DM-7102. I suspect that everything I know about it is available on those tickets (and possible ones linked from them).

python/lsst/drp/tasks/wcsfit.py
```diff
@@ -0,0 +1,1145 @@
...
+                datasetType,
+                collections=collections,
+                dataId=visit_data_id,
+                findFirst=True,
+            ).expanded()
+        )
+    yield from refs
+
+
+def make_ref_covariance_matrix(refCat, inputUnit=u.radian, outputCoordUnit=u.marcsec, outputPMUnit=u.marcsec):
```
erykoff, line 84: These utility functions aren't in the exported `__all__`.  I assume this is intentional, and these are just utility functions used by the task?  If so, I wonder if it would make sense to rename them all with a leading underscore to make it super clear these aren't supposed to be public.

erykoff (commented):
This looks very good to me, with some small comments, and you should resolve conversations that you think are good.

One meta-question that I think has been brought up before is the name of the task.  `WcsFitTask` is somewhat generic, and doesn't make it clear from the name that this is based on the `gbdes` code.  Have you considered putting this in the task name?  (At the minimum the documentation overview should say in the opening paragraph that this uses `gbdes` to do the fitting, perhaps with a link to the relevant DES paper?). The reason I think the name is important is that we (well, myself at least) are publicly talking about moving to "using gbdes for astrometric fitting", but then there's nothing in the pipeline definition that makes that apparent.

python/lsst/drp/tasks/wcsfit.py
parejkoj: I believe the expected frames are the ones documented here: https://pipelines.lsst.io/modules/lsst.afw.cameraGeom/cameraGeom.html

tests/test_wcsfit.py
parejkoj: Re-asking my question above, as it probably got lost in the noise: Where does the number of detectors and their dimensions (that go into the wcsfit.Instrument object) come from? You should be getting that from the Camera object.

python/lsst/drp/tasks/wcsfit.py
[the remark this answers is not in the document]
cmsaunders: How about using `lsst.daf.butler.formatters.numpy_dict_to_arrow` since the original format is a dict of numpy arrays? And would the storage class be "ArrowNumpy" in that case?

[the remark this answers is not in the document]
erykoff: If you manually convert from numpy dict to arrow, the storage class would be `ArrowTable` in that case since you have converted it to an arrow table already and are storing it as an arrow table.  (There actually isn't a `NumpyDictArrow` or equivalent storage class which would allow you to store the numpy dict directly).  Anyway, doing the `numpy_dict_to_arrow` conversion and storing as an `ArrowTable` is a good idea and would probably be the most direct way of doing this.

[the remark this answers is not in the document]
timj: It's `lsst.daf.butler.formatters.parquet.numpy_dict_to_arrow`. If we are using these functions in user code and not internally in storage class converters then I'm a bit worried about them being in the wrong place in daf_butler.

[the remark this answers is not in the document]
erykoff: @timj This is a bit off-topic, the numpy dict conversion code is used internally, it just isn't exposed as a first-class storage class.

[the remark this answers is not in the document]
timj: Yes, I know we are using them internally, my comment was that the helper functions aren't intended to be public API for task code.

python/lsst/drp/tasks/wcsfit.py
cmsaunders: The names here, if I am reading it correctly, don't match the frame names in `jointcal`, so I don't think this can be right. I've been testing `WCSFit` in the DRP-RC2 pipeline, and the downstream tasks seem to work with this configuration of the wcs.

tests/test_wcsfit.py
cmsaunders: The number and dimensions of the detectors comes from the `visitSummary` tables. The layout comes from the input wcs, which is also in the `visitSummary` tables.

python/lsst/drp/tasks/wcsfit.py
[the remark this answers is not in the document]
erykoff: I have filed https://jira.lsstcorp.org/browse/DM-37279 and will add a direct `ArrowNumpyDict` storage class which should be used here when that's ready (soon!).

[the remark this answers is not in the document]
erykoff: A simpler alternative that doesn't require that ticket: if you don't have any string columns, then you can easily use `pyarrow.Table.from_dict()` to convert to an arrow table, and then use `ArrowTable` storage class.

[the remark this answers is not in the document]
cmsaunders: Right, I think we talked about doing this a while ago. This is the approach I'm going to take.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -0,0 +1,1150 @@
...
+    )
+    sourceFluxType = pexConfig.Field(
+        dtype=str,
+        doc="Source flux field to use in source selection and to get fluxes from the catalog.",
+        default='apFlux_12_0'
+    )
+    systematicError = pexConfig.Field(
+        dtype=float,
+        doc=("Systematic error padding added in quadrature for the science catalogs (marcsec). The default"
+             "value is equivalent to 0.02 pixels for HSC"),
```
erykoff, line 302: Missing period at end.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -0,0 +1,1150 @@
...
+    )
+    referenceSystematicError = pexConfig.Field(
+        dtype=float,
+        doc="Systematic error padding added in quadrature for the reference catalog (marcsec).",
+        default=0.0
+    )
+    modelComponents = pexConfig.ListField(
+        dtype=str,
+        doc=("List of mappings to apply to transform from pixels to sky, in order of their application."
+             "Supported options are 'INSTRUMENT/DEVICE' and 'EXPOSURE'"),
```
erykoff, line 313: Missing period at end.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -0,0 +1,1150 @@
...
+        default=["INSTRUMENT/DEVICE", "EXPOSURE"]
+    )
+    deviceModel = pexConfig.ListField(
+        dtype=str,
+        doc=("List of mappings to apply to transform from detector pixels to intermediate frame. Map names"
+             "should match the format 'BAND/DEVICE/<map name>'"),
```
erykoff, line 319: Missing period at end.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -0,0 +1,1150 @@
...
+        default=["BAND/DEVICE/poly"]
+    )
+    exposureModel = pexConfig.ListField(
+        dtype=str,
+        doc=("List of mappings to apply to transform from intermediate frame to sky coordinates. Map names"
+             "should match the format 'EXPOSURE/<map name>'"),
```
erykoff, line 325: Missing period at end.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -0,0 +1,1150 @@
...
+                    ]
+        self.sourceSelector["science"].flags.bad = badFlags
+
+    def validate(self):
+        super().validate()
+
+        # Check if all components of the device and exposure models are
+        # supported.
+        for component in self.deviceModel:
+            if not (('poly' in component.lower()) or ('identity' in component.lower())):
```
erykoff, line 379: I'm going to be "that guy" who mentions that this file has a mix of single-quotes and double-quotes.  Now is the time to pick one.  (I personally like single quotes, but black prefers double quotes; I leave it up to you to choose which format for these files.)

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -0,0 +1,1150 @@
...
+                List of exposure catalogs (one per visit) with the WCS for each
+                detector set by the new fitted WCS.
+            ``fitModel`` : `wcsfit.WCSFit`
+                Model-fitting object with final model parameters.
+            ``outputCatalog`` : `pyarrow.Table`
+                Catalog with fit residuals of all sources used.
+        """
+        self.log.info("Gathering instrument, exposure, and field info")
+        # Set up an instrument object
+        instrument = wcsfit.Instrument(instrumentName)
```
erykoff, line 460: If instrument is optional, what happens if you call this with `None`?

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -0,0 +1,1150 @@
...
+                                        detectorSources[goodInds]['x'].to_list(),
+                                        detectorSources[goodInds]['y'].to_list(),
+                                        np.arange(goodInds.sum()))
+
+        associations.sortMatches(fieldIndex, minMatches=self.config.minMatches,
+                                 allowSelfMatches=self.config.allowSelfMatches)
+
+        return sourceIndices, columns
+
+    def make_YAML(self, inputVisitSummary, inputFile=None):
```
erykoff, line 886: I think this should be called `make_yaml` not `make_YAML`.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -0,0 +1,1150 @@
...
+    def _add_objects(self, wcsf, inputCatalogRefs, sourceIndices, extensionInfo, columns):
+        """Add science sources to the wcsfit.WCSFit object.
+
+        Parameters
+        ----------
+        wcsf : `wcsfit.WCSFit`
+            WCS-fitting object.
+        inputCatalogRefs : `list`
+            List of DeferredDatasetHandles pointing to visit-level source.
+            tables
```
erykoff, line 959: The period comes on the previous line, and `tables` is sitting sadly alone.

erykoff (approved):
Some minor formatting things to be cleaned up; nothing major.  One question I have is if gbdes needs specific support for every instrument, or is the name just used as a marker?

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: Ok, I kept double quotes for documentation, but standardized to single quotes otherwise.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: Addressed in docs.

PR #11, opened 2023-04-17
DM-37703: Remove DatasetRef from test

timj:
The test does not require an unresolved ref.

Requires lsst/daf_butler#821


andy-slac (approved):
Looks good.

PR #19, opened 2023-07-06
DM-35901: Use the new lsst.utils.tests.ImportTestCase class for import test

timj:
Requires lsst/utils#165


tests/test_import.py
```diff
@@ -1,4 +1,4 @@
-# This file is part of drp_tasks.
+# This file is part of pipe_tasks.
```
erykoff, line 1: You changed all the references in this file from drp_tasks (correct) to pipe_tasks (not correct).  (I assume an errant cut-and-paste).

timj: Fixed.

PR #22, opened 2023-08-25
DM-40513: Add first-draft metadetection PipelineTask.

TallJimbo:


python/lsst/drp/tasks/metadetection_shear.py
```diff
@@ -0,0 +1,286 @@
...
+        Returns
+        -------
+        results : `lsst.pipe.base.Struct`
+            Structure with the following attributes:
+
+            - ``object_catalog`` [ `pyarrow.Table` ]: the output object
+              catalog for the patch, with schema equal to `object_schema`.
+        """
+        single_cell_tables: list[pa.Table] = []
+        for single_cell_coadds in zip(*patch_coadds, strict=True):
```
esheldon, line 256: This is a loop over bands right?  We will be making a combined coadd over bands for detection, but then process the bands simultaneously to get the shear info, fluxes etc.

python/lsst/drp/tasks/metadetection_shear.py
```diff
@@ -0,0 +1,286 @@
...
+            single_cell_tables.append(self.process_cell(single_cell_coadds))
+        # TODO: if we need to do any cell-overlap-region deduplication here
+        # (instead of purely in science analysis code), this is where'd it'd
+        # happen.
+        return Struct(
```
esheldon, line 261: Does this struct have to match the schema returned by  make_object_schema?

python/lsst/drp/tasks/metadetection_shear.py
```diff
@@ -0,0 +1,286 @@
...
+            Per-band, per-cell coadds, in the order specified by
+            `MetadetectionShearConfig.required_bands`.
+
+        Returns
+        -------
+        object_catalog : `pyarrow.Table`
+            Output object catalog for the cell, with schema equal to
+            `object_schema`.
+        """
+        rows: list[dict[str, Any]] = []
```
esheldon, line 281: Must it be a list/dict or can it be a numpy array with fields?  e.g. `np.zeros(nrows, dtype=dtype)`?

python/lsst/drp/tasks/metadetection_shear.py
```diff
@@ -0,0 +1,286 @@
...
+        doc="Per-band deep coadds.",
+        multiple=True,
+        dimensions={"patch", "band"},
+    )
+
+    # TODO: If there are image-like or other non-catalog output products (e.g.
+    # detection masks), add them here.
+
+    object_catalog = cT.Output(
+        "ShearObject",
```
esheldon, line 64: Does this name matter?  If so should this be more specific like "MDetShearObject" ?

TallJimbo: It does matter, but for at least DR1 I think "ShearObject" makes sense as there will be only one shear catalog table and that's what we've put in the DPDD.  @arunkannawadi may ultimately have been right that we should have put a more specific name in the DPDD, too, but I'm comfortable revisiting that for DR2+ if/when we decide to run another algorithm alongside.

python/lsst/drp/tasks/metadetection_shear.py
TallJimbo: This is a loop over individual cells, which each `single_cell_coadds` object a sequence of per-band single-cell coadds.

I figured we'd coadd across bands inside `process_cell`, one cell at a time.  I was assuming there'd be no reason to coadd all cells across bands up front and then process both the multi-band coadd and the per-band coadds for each cell more later.

python/lsst/drp/tasks/metadetection_shear.py
TallJimbo: The `object_catalog` attribute of the struct needs to be a pyarrow table that matches that schema.  The attributes of the `Struct` itself correspond to the `cT.Output` connections in `MetadetectionShearConnections`, of which there is only one right now.

python/lsst/drp/tasks/metadetection_shear.py
TallJimbo: I'm pretty sure numpy array with fields can work too, because @erykoff has a function that can convert from that to `pyarrow.Table` we can import and use.

python/lsst/drp/tasks/metadetection_shear.py
esheldon: I won't be surprised if Xiangchong Li has a code ready by then

python/lsst/drp/tasks/metadetection_shear.py
erykoff: I do!  What I don't have is a way of connecting that with the arrow schema metadata, but once we have an idea of what functionality we need I can add that.

python/lsst/drp/tasks/metadetection_shear.py
esheldon: Ah, I see now there is a star on patch_coadds.  So this means that patch_coadds is actuallly a sequence of sequences?  Where can I find the definition of MultiCellCoadd?

TallJimbo: Oops, I wrote this thinking that `MultipleCellCoadd` was iterable, but it isn't (you have to iterate over `.cells`) so this was actually wrong (just pushed a fix).

Here's the definition of `MultipleCellCoadd`: https://github.com/lsst/cell_coadds/blob/main/python/lsst/cell_coadds/_multiple_cell_coadd.py

And here's the definition of `SingleCellCoadd`: https://github.com/lsst/cell_coadds/blob/main/python/lsst/cell_coadds/_single_cell_coadd.py

Those both refer to [`ImagePlanes`](https://github.com/lsst/cell_coadds/blob/main/python/lsst/cell_coadds/_image_planes.py) and [`CommonComponents`](https://github.com/lsst/cell_coadds/blob/main/python/lsst/cell_coadds/_common_components.py) from the same package.

esheldon: Thanks.

Is there an existing way to coadd these individual band SingleCellCoadds or should we extract the masked image and do it ourselves?

How do extract the image?

esheldon: (assuming we should not rely on hidden `_outer` attribute

esheldon: whoops, I see there is a property for outer

TallJimbo: Go ahead and extract the outer masked image or individual outer single-plane images and do it yourself; it might make sense to move that into a method on `SingleCellCoadd` for reuse at some point, but we can look into that once you've defined the behavior you want for the combination.

PR #34, opened 2023-11-01
DM-41512: Add check for no good sources in finalizeCharacterization

laurenam:


python/lsst/drp/tasks/update_visit_summary.py
```diff
@@ -706,6 +706,8 @@ def run(
                     psf = None
                 output_record.setPsf(psf)
                 sources = psf_star_catalog[psf_star_catalog["detector"] == detector_id]
+                if len(sources) < 1:
```
erykoff, line 709: I would normally do `len(sources) == 0`, but I suppose this would do the same.

erykoff (approved):

python/lsst/drp/tasks/update_visit_summary.py
timj: or
```python
if not sources:
    ...
```

PR #36, opened 2023-11-27
DM-41634: Switch initial WCS to AST

cmsaunders:


TallJimbo (approved):
Changes look fine, but please squash the commits.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -708,7 +708,12 @@ def _get_exposure_info(
                 extensionType.append("SCIENCE")
 
                 wcs = row.getWcs()
-                wcss.append(_get_wcs_from_sip(wcs))
```
timj, line 711: I think you can delete the `_get_wcs_from_sip` function completely now.

tests/test_gbdesAstrometricFit.py
```diff
@@ -409,15 +409,8 @@ def test_get_exposure_info(self):
...
-                # One WCS is in SIP and the other is TPV. The pixel-to-sky
-                # conversion is not exactly the same but should be close.
-                # TODO: sip_tpv + astropy.wcs.WCS gets a better result here,
-                # particularly for detector # >= 100. See if we can improve/if
-                # improving is necessary. Check if matching in corner detectors
-                # is ok.
-                rtol = 1e-3 if (detector >= 100) else 1e-5
-                np.testing.assert_allclose(calexpra, newRAdeg, rtol=rtol)
-                np.testing.assert_allclose(calexpdec, newDecdeg, rtol=rtol)
+                np.testing.assert_allclose(calexpra, newRAdeg)
```
timj, line 412: It's nice to be able to simplify this test so it no longer has to account for edge detectors.

timj (approved):

tests/test_gbdesAstrometricFit.py
cmsaunders: Yes, I was quite happy about that, and that the transformations for the other detectors are more accurate too.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: I'm reluctant to remove this, as it was a pain to get it working properly and I'm not 100% certain it won't be needed in the future.

PR #37, opened 2023-12-04
DM-41635: Add option to output model parameters and cov

cmsaunders:


python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -218,10 +218,22 @@ class GbdesAstrometricFitConnections(
         storageClass="ArrowNumpyDict",
         dimensions=("instrument", "skymap", "tract", "physical_filter"),
     )
+    modelParams = pipeBase.connectionTypes.Output(
+        doc="WCS parameter covariance.",
+        name="gbdesAstrometricFit_modelParams",
+        storageClass="ArrowNumpyDict",
```
taranu, line 224: TIL...

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -302,6 +314,14 @@ class GbdesAstrometricFitConfig(
         doc="Set the random seed for selecting data points to reserve from the fit for validation.",
         default=1234,
     )
+    saveModelParams = pexConfig.Field(
+        dtype=bool,
+        doc=(
+            "Save the parameters and covariance of the WCS model. Default to "
+            "false because this can be very large."
```
taranu, line 321: I don't know if these needs to be added to the docstring necessarily but how many parameters is typical? Or is the possible range very large?

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -517,9 +539,17 @@ def run(
         outputWCSs = self._make_outputs(wcsf, inputVisitSummaries, exposureInfo)
         outputCatalog = wcsf.getOutputCatalog()
         starCatalog = wcsf.getStarCatalog()
+        if self.config.saveModelParams:
```
taranu, line 542: I prefer one-line ternary `modelParams = self._output_model_params(wcsf) if self.config.saveModelParams else None`, but your taste may vary.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -1233,3 +1263,47 @@ def _make_outputs(self, wcsf, visitSummaryTables, exposureInfo):
             catalogs[visit] = catalog
 
         return catalogs
+
+    def _output_model_params(self, wcsf):
+        """Get the WCS model parameters and covariance and convert to a
+        dictionary that will be readable as a pandas dataframe.
```
taranu, line 1269: I'm assuming it returns a numpy ndarray by default (with column/row names?), so while I guess you can convert it to a dataframe why would you want to?

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -1233,3 +1263,47 @@ def _make_outputs(self, wcsf, visitSummaryTables, exposureInfo):
             catalogs[visit] = catalog
 
         return catalogs
+
+    def _output_model_params(self, wcsf):
```
taranu, line 1267: I'd call this `_compute_model_params` if it's returning them and not actually doing output itself.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -1233,3 +1263,47 @@ def _make_outputs(self, wcsf, visitSummaryTables, exposureInfo):
...
+
+        Returns
+        -------
+        modelParams : `dict`
+            Parameters and covariance of the best-fit WCS model.
+        """
+        modelParamDict = wcsf.mapCollection.getParamDict()
+        modelCovariance = wcsf.getModelCovariance()
+
+        modelParams = {"mapName": [], "coordinate": [], "parameter": [], "coefficientNumber": []}
```
taranu, line 1284: It doesn't particularly matter with a short list but this could be a dict comp or defaultdict.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -1233,3 +1263,47 @@ def _make_outputs(self, wcsf, visitSummaryTables, exposureInfo):
...
+            modelParams["coefficientNumber"].extend(np.arange(nCoordCoeffs))
+            modelParams["coefficientNumber"].extend(np.arange(nCoordCoeffs))
+
+            for p in range(nCoeffs):
+                if p < nCoordCoeffs:
+                    coord = "x"
+                else:
+                    coord = "y"
+                modelParams[f"{mapName}_{coord}_{p}_cov"] = modelCovariance[i]
+                i += 1
```
taranu, line 1303: Was this just for debugging? The value never seems to be read.

tests/test_gbdesAstrometricFit.py
```diff
@@ -491,6 +492,16 @@ def test_make_outputs(self):
...
 
+    def test_output_model_params(self):
+        """Test the optional model parameters and covariance output."""
+        modelParams = pd.DataFrame(self.outputs.modelParams)
+        # Check that DataFrame is the expected size.
+        shape = modelParams.shape
+        self.assertEqual(shape[0] + 4, shape[1])
+        # Check that covariance matrix is symmetric.
+        covariance = (modelParams.iloc[:, 4:]).to_numpy()
+        np.testing.assert_array_almost_equal(covariance, covariance.T)
```
taranu, line 503: Why almost? Shouldn't they be identical, otherwise it's not a covariance matrix? Note that the default for this check is 1e-6 precision and that's way more than typical roundoff error.

taranu (commented):
I have some questions and pedantic comments, except for the one about checking covariance matrix symmetry which is actually important.

tests/test_gbdesAstrometricFit.py
cmsaunders: In the test, the max absolute difference is 2e-19, but the max relative difference is 2e-6, so it doesn't pass (for 2 out of 10614564 elements).

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: The `i`? It's in the line above.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: How about just `_get_model_params`?

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: The `ArrowNumpyDict` is a dictionary of numpy arrays. I guess I could just say that it can be converted to a table of some sort. Turning it into a dataframe is helpful for me for keeping track of what the parameters are, but maybe not the most efficient path.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: The range is really large because the number of model coefficients depends on the polynomial order for the two mappings, the number of detectors, and the number of visits. For a tract of HSC RC2, it's ~5000 parameters.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: About ArrowNumpyDicts?

python/lsst/drp/tasks/gbdesAstrometricFit.py
taranu: If it's just copying values and not really computing anything, ok.

python/lsst/drp/tasks/gbdesAstrometricFit.py
taranu: Yep, I don't see any other uses in the stack yet (probably because most devs also don't know it exists).

tests/test_gbdesAstrometricFit.py
taranu: You might want to use [assert_allclose](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.htm) instead then, since that allow for setting separate absolute and relative tolerances.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: Yeah, now that I'm thinking about it, this task's outputs are what triggered Eli adding this storage class.

PR #38, opened 2023-12-04
DM-40652: Raise error when data is insufficient

cmsaunders:


python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -436,6 +436,11 @@ def run(
             ``outputCatalog`` : `pyarrow.Table`
                 Catalog with fit residuals of all sources used.
         """
+        if (len(inputVisitSummaries) == 1) and (self.config.deviceModel and self.config.exposureModel):
+            raise RuntimeError(
+                "More than one exposure is necessary to break the degeneracy between the"
```
arunkannawadi, line 441:
```suggestion
                "More than one exposure is necessary to break the degeneracy between the "
```

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -436,6 +436,11 @@ def run(
...
+                " device model and the exposure model."
```
arunkannawadi, line 442:
```suggestion
                "device model and the exposure model."
```

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -436,6 +436,11 @@ def run(
             ``outputCatalog`` : `pyarrow.Table`
                 Catalog with fit residuals of all sources used.
         """
+        if (len(inputVisitSummaries) == 1) and (self.config.deviceModel and self.config.exposureModel):
```
arunkannawadi, line 439: You could skip the set of parenthesis towards the end of the line.
https://developer.lsst.io/python/style.html#the-minimum-number-of-parentheses-needed-for-correctness-and-readability-should-be-used

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -508,9 +514,16 @@ def run(
         self._add_objects(wcsf, inputCatalogRefs, sourceIndices, extensionInfo, usedColumns)
         self._add_ref_objects(wcsf, refObjects, refCovariance, extensionInfo)
 
+        # There must be at least as many sources per visit as the number of
+        # free parameters in the per-visit mapping. Set minFitExposures to be
+        # the number of free parameters, so that visits with fewer visits are
+        # dropped.
+        nCoeffVisitModel = int(((2 * self.config.exposurePolyOrder + 3) ** 2 - 1) / 8)
```
arunkannawadi, line 521: Could you briefly comment on how this calculation comes about? Even better would be to make this a utility function and comment it as a docstring, since you use this calculation at least twice in this code.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -1199,6 +1254,12 @@ def _make_outputs(self, wcsf, visitSummaryTables, exposureInfo):
         for v, visitSummary in enumerate(visitSummaryTables):
             visit = visitSummary[0]["visit"]
 
+            visitMap = wcsf.mapCollection.orderAtoms(f"{visit}")[0]
+            visitMapType = wcsf.mapCollection.getMapType(visitMap)
+            if (visitMap not in mapParams) and (visitMapType != "Identity"):
+                self.log.warning(f"Visit {visit} was dropped because of an insufficient number of sources.")
```
arunkannawadi, line 1260:
```suggestion
                self.log.warning("Visit %d was dropped because of an insufficient number of sources.", visit)
```

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -948,6 +962,47 @@ class `wcsfit.FoFClass`, associating them into matches as you go.
 
         return sourceIndices, columns
 
+    def _check_degeneracies(self, associations, extensionInfo):
```
arunkannawadi, line 965: It looks like you could decorate it as a staticmethod as it doesn't use `self` in the body.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -948,6 +962,47 @@ class `wcsfit.FoFClass`, associating them into matches as you go.
...
+            unconstrainedDetectors = []
+            for detector in np.unique(extensionInfo.detector):
+                numSources = (whichDetector == detector).sum()
+                if numSources < nCoeffDetectorModel:
+                    unconstrainedDetectors.append(str(detector))
+
+            if unconstrainedDetectors:
+                raise RuntimeError(
+                    "The model is not constrained. The following detectors do not have enough "
+                    f"sources ({nCoeffDetectorModel} required): " + ", ".join(unconstrainedDetectors)
```
arunkannawadi, line 1003: Because this is within parenthesis, you could drop the `+` and the strings would concatenate anyway.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: True, I was thinking about `self.config.deviceModel` and `self.config.exposureModel` going together, but I realize it's logically the same without parentheses.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: Can you explain this suggestion? Is there an issue with f-strings in logging, or are you otherwise concerned about formatting of `visit`?

timj: f-strings in loggers are frowned upon and non-standard where deferred formatting is the norm. For warnings it doesn't really matter because we can be very sure that the warning will always be issued (whereas for debugs you definitely don't want to use f-strings).

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: Ok, I decided to make utility functions for this and for the inverse operation.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: It does use a config parameter actually, so I need `self`.

python/lsst/drp/tasks/gbdesAstrometricFit.py
arunkannawadi: What Tim said.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -121,6 +121,49 @@ def _make_ref_covariance_matrix(
...
+    ----------
+    degree : `int`
+        Degree of the polynomial in question.
+
+    Returns
+    -------
+    nCoeffs : `int`
+        Number of coefficients for the polynomial in question.
+    """
+    nCoeffs = int(((2 * degree + 3) ** 2 - 1) / 8)
```
arunkannawadi, line 142: This formula wasn't obvious but makes sense now after reading the docstring. Why not just `(degree+2)*(degree+1)/2` though? It doesn't seem like the simplest form of the quantity you want to compute.

arunkannawadi (approved):

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: Yeah, I don't remember why I didn't simplify this when I first implemented it. I think my idea was to preserve the structure of the formula, but I don't think it's necessary.

PR #40, opened 2023-12-15
DM-41490: Recover WCS for input with astrometry failures

cmsaunders:


python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -999,9 +1009,14 @@ class `wcsfit.FoFClass`, associating them into matches as you go.
...
 
                 isStar = np.ones(goodInds.sum())
-                extensionIndex = np.flatnonzero(
+                findExtension = np.flatnonzero(
                     (extensionInfo.visit == visit) & (extensionInfo.detector == detector)
-                )[0]
+                )
+                if len(findExtension) == 0:
+                    # This extension does not have information necessary for
+                    # fit. Skip these detections.
```
parejkoj, line 1017: I'm not following this comment (I don't know what an extension is in this context). By "these detections", do you mean "this detector/visit"?

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -1164,9 +1181,15 @@ def _add_objects(self, wcsf, inputCatalogRefs, sourceIndices, extensionInfo, col
...
                 detectorSources = inputCatalog[inputCatalog["detector"] == detector]
 
-                extensionIndex = np.flatnonzero(
+                findExtension = np.flatnonzero(
                     (extensionInfo.visit == visit) & (extensionInfo.detector == detector)
-                )[0]
+                )
+                if len(findExtension) == 0:
+                    # This extension does not have information necessary for
+                    # fit. Skip these detections.
```
parejkoj, line 1189: Same comment as above, with the aded question of why this is duplicated here?

parejkoj (commented):
There's no test of this functionality: you should be able to make a trivial visitSummary that has a None (or missing) WCS for one detector and test with that.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -1335,8 +1358,23 @@ def _make_outputs(self, wcsf, visitSummaryTables, exposureInfo):
...
                 mapName = f"{visit}/{detector}"
-
-                mapElements = wcsf.mapCollection.orderAtoms(f"{mapName}/base")
+                if mapName in wcsf.mapCollection.allMapNames():
+                    mapElements = wcsf.mapCollection.orderAtoms(f"{mapName}/base")
+                else:
+                    # This extension was not fit, but try to recover WCS
+                    genericElements = mapTemplate["EXPOSURE/DEVICE/base"]["Elements"]
+                    mapElements = []
+                    instrument = visitSummary[0].getVisitInfo().getInstrumentLabel()
```
parejkoj, line 1367: You should be able to do `summary.visitInfo.instrumentLabel`, instead of using getters.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -1335,8 +1358,23 @@ def _make_outputs(self, wcsf, visitSummaryTables, exposureInfo):
...
+                    for component in genericElements:
+                        elements = mapTemplate[component]["Elements"]
+                        for element in elements:
+                            for generic, specific in {
+                                "BAND": instrument,
+                                "EXPOSURE": visit,
+                                "DEVICE": detector,
+                            }.items():
+                                element = element.replace(generic, str(specific))
```
parejkoj, line 1376: This seems unnecessarily complicated. Why not just do these three things one at a time, so it's easier to follow? Making it a loop like this is just extra lines and harder to debug.

```
element = element.replace("BAND", instrument)
element = element.replace("EXPOSURE", visit)
element = element.replace"DEVICE", detector)
```

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -1335,8 +1358,23 @@ def _make_outputs(self, wcsf, visitSummaryTables, exposureInfo):
...
+                else:
+                    # This extension was not fit, but try to recover WCS
+                    genericElements = mapTemplate["EXPOSURE/DEVICE/base"]["Elements"]
+                    mapElements = []
+                    instrument = visitSummary[0].getVisitInfo().getInstrumentLabel()
+                    for component in genericElements:
+                        elements = mapTemplate[component]["Elements"]
+                        for element in elements:
+                            for generic, specific in {
+                                "BAND": instrument,
```
parejkoj, line 1372: Is `instrument` really a thing to substitute for `band`?

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -1335,8 +1358,23 @@ def _make_outputs(self, wcsf, visitSummaryTables, exposureInfo):
...
-
-                mapElements = wcsf.mapCollection.orderAtoms(f"{mapName}/base")
+                if mapName in wcsf.mapCollection.allMapNames():
+                    mapElements = wcsf.mapCollection.orderAtoms(f"{mapName}/base")
+                else:
+                    # This extension was not fit, but try to recover WCS
+                    genericElements = mapTemplate["EXPOSURE/DEVICE/base"]["Elements"]
+                    mapElements = []
+                    instrument = visitSummary[0].getVisitInfo().getInstrumentLabel()
+                    for component in genericElements:
```
parejkoj, line 1368: I really don't understand what this nested set of loops is supposed to do. I assume it's a gbdes internals thing?

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -1335,8 +1358,23 @@ def _make_outputs(self, wcsf, visitSummaryTables, exposureInfo):
 
             for d, detector in enumerate(visitSummary["id"]):
                 mapName = f"{visit}/{detector}"
-
-                mapElements = wcsf.mapCollection.orderAtoms(f"{mapName}/base")
+                if mapName in wcsf.mapCollection.allMapNames():
+                    mapElements = wcsf.mapCollection.orderAtoms(f"{mapName}/base")
+                else:
+                    # This extension was not fit, but try to recover WCS
```
parejkoj, line 1364: Assuming "extension" means "detector/visit": if a detector/visit wasn't fit, how could there be a WCS?

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -738,6 +739,23 @@ def _get_exposure_info(
...
                 detector = row["id"]
+
+                wcs = row.getWcs()
+                if wcs is None:
+                    self.log.warning(
+                        "WCS is None for visit %d, detector %d: this extension will be dropped.",
+                        visit,
+                        detector,
+                    )
+                    continue
```
parejkoj, line 750: Should this be a warning, or an info-level message? Also, saying "extension" in the log is unhelpful for anyone who doesn't know the gbdes internals.

parejkoj (changes requested):
Sorry, I clicked the button too soon earlier.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: `summary.visitInfo` doesn't work, but I can do `visitSummary[0].getVisitInfo().instrumentLabel`.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: It probably shouldn't be, but these lines are trying to replicate how `gbdes` names mappings, and the instrument name is what ends up here (probably due to this line: https://github.com/lsst/gbdes/blob/4efbc7097dc222cb35d22f8aed79572413aa1a97/include/Instrument.h#L21). A near-term item on my to-do list is to switch to fitting multiple bands at once, so this will probably be addressed there.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: This is duplicated because the sources get loaded two times--once when the source association is done, and again when they are added to the model fitting object.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: "extension" is the `gbdes` term for one detector from one visit, and it's used throughout the `gbdes` task. I can change this line to "Skip the detections from this detector for this visit" to make it more clear.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: Sort of...this loop goes through the generic definitions for the mappings that take you from pixels to sky and uses those to get the names for the maps for a specific extension (i.e. visit/detector). I spent a lot of time trying to get `gbdes` to do this itself, but couldn't manage it, so this is my replication of how `gbdes` makes the names.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: Yes, "extension" means "detector/visit". The detector part of the model is fixed by the sources from this detector on other visits, and the visit part of the model is fixed using the other detectors for the visit. Actually, this makes me realize I need to add a check so that this doesn't fail when there is only one detector.

cmsaunders: Scratch the part about what happens when there is only one detector. That will be caught at line 1354.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: I think a warning is right (from the dev-guide: "WARNING: for conditions that may indicate a problem but that allow continued execution"). I think there should be a note of caution for the resulting WCS. 
I changed the message to say "...this extension (visit/detector) will be dropped."

cmsaunders:
Thanks for the review @parejkoj. I'll add a unit test and update the PR.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -1335,8 +1361,25 @@ def _make_outputs(self, wcsf, visitSummaryTables, exposureInfo):
...
+                    # this visit.
+                    genericElements = mapTemplate["EXPOSURE/DEVICE/base"]["Elements"]
+                    mapElements = []
+                    instrument = visitSummary[0].getVisitInfo().instrumentLabel
+                    # Go through the generic map components to build the names
+                    # of the specific maps for this extension.
+                    for component in genericElements:
+                        elements = mapTemplate[component]["Elements"]
+                        for element in elements:
+                            element = element.replace("BAND", str(instrument))
```
parejkoj, line 1379: I think you need a comment here as to why band and instrument are being mixed.

tests/test_gbdesAstrometricFit.py
```diff
@@ -539,6 +539,31 @@ def test_run(self):
...
+        outputs = self.task.run(
+            self.inputCatalogRefs,
+            inputVisitSummary,
+            instrumentName=self.instrumentName,
+            refEpoch=self.refEpoch,
+            refObjectLoader=self.refObjectLoader,
+        )
+
+        # Check that the fit WCS for the extension with input WCS=None returns
+        # finite sky values.
```
parejkoj, line 562: Should it return finite sky values, or correct sky values? Is the test dataset not large enough to get a reasonable fit in this case?

parejkoj (commented):
You really should pull the contents of some of these for-loops into separate functions. That will make it easier to parse, and at least in one case here, you won't have duplicated code.

python/lsst/drp/tasks/gbdesAstrometricFit.py
parejkoj: Do you have a way to flag the WCSs that are "reconstructed" in this case in the output, to warn users? A warning message may be appropriate here, but it's not sufficient to tell people that the output WCS is potentially unreliable.

cmsaunders: No, neither the WCSs nor the visitSummary tables have a flag for WCS quality. The WCS fit will have higher uncertainty, as it would if there were very few stars to fit, and Eli and I have talked a bit about how to propagate this information.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -1164,9 +1183,16 @@ def _add_objects(self, wcsf, inputCatalogRefs, sourceIndices, extensionInfo, col
...
                     (extensionInfo.visit == visit) & (extensionInfo.detector == detector)
-                )[0]
+                )
+                if len(findExtension) == 0:
+                    # This extension does not have information necessary for
+                    # fit. Skip the detections from this detector for this
+                    # visit.
+                    continue
+                extensionIndex = findExtension[0]
+
```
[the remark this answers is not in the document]
cmsaunders: I looked at this for a while and decided that the two loops are different enough that a function is not going to improve clarity much. I did implement a `_find_extension_index()` method for a part of the loop that is repeated.

tests/test_gbdesAstrometricFit.py
cmsaunders: The test data is simulated without the missing WCS, so we don't expect to be able to recover the same input in this test. It's possible to do a whole new simulation, but I don't think it's worthwhile in this case.

parejkoj (approved):
Thanks for the cleanups.

PR #44, opened 2024-02-14
DM-42663: Change default sourceSelector to new moments based selector in gbdes

erykoff:


tests/test_gbdesAstrometricFit.py
```diff
@@ -340,6 +340,7 @@ def _make_sourceCat(cls, starIds, starRas, starDecs, trueWCSs, inScienceFraction
                     "hsmPsfMoments_flag",
                     "apFlux_12_0_flag",
                     "extendedness",
+                    "sizeExtendedness",
```
parejkoj, line 343: Any reason you have both of these here?

TallJimbo (approved):

tests/test_gbdesAstrometricFit.py
arunkannawadi: We still want to propagate the what I call "flux extendedness" values, but switch from using one to another. At some future point, I'd like to make `slot_extendedness` that can point to either, but not in the near future.

arunkannawadi (approved):

PR #45, opened 2024-03-07
DM-42431: Enable coadds from warps with absolute calibration

arunkannawadi:
This PR allows to skip scaling to a common zero point by default, since the warps from `MakeDirectWarp` task produce warps with pixel units in nJy.

### DIFF
--- a/python/lsst/drp/tasks/assemble_cell_coadd.py
+++ b/python/lsst/drp/tasks/assemble_cell_coadd.py
@@ -335,6 +335,17 @@ class AssembleCellCoaddTask(PipelineTask):
 
         cells: list[SingleCellCoadd] = []
         for cellInfo in skyInfo.patchInfo:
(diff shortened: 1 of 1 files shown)


arunkannawadi:
This is a new PR to an already-reviewed ticket.

TallJimbo (approved):
Thanks for the quick fix - I saw an error report from @mirarenee about what I'm guessing was this issue yesterday and hadn't had time to follow-up yet.

arunkannawadi:
Yup, this was one part of the issue and the other one should be fixed in `MakeDirectWarp` itself. I had not tested the new warping task with the new coaddition task yet, and I'm glad Miranda caught this incompatibility so we could fix it before merging to `main`.

arunkannawadi:
Closing this PR without merging. The changes here will instead appear in DM-43515 and DM-43516.

PR #47, opened 2024-03-29
DM-43516: Pass ObservationalIdentifiers to SingleCellCoadd

arunkannawadi:


python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -359,7 +403,7 @@ def run(self, inputWarps, skyInfo, **kwargs):
                 outer=image_planes,
                 psf=cell_coadd_psf.computeKernelImage(cell_coadd_psf.getAveragePosition()),
                 inner_bbox=cellInfo.inner_bbox,
-                inputs=None,  # TODO
+                inputs=frozenset(observation_identifiers_gc[cellInfo.index]),
```
TallJimbo, line 406: If we use a `frozenset` or `set` here we need to make sure we sort on save to avoid making the persisted order dependent on salted hashing.  Might be better to remove duplicates then sort it and make a `tuple` right here.

TallJimbo: Also, we really don't want to be using `getAveragePosition` two lines above.  That is the average position of all the stars that went into the PSF model, and it's not necessarily going to be anywhere near this cell.

TallJimbo (approved):

python/lsst/drp/tasks/assemble_cell_coadd.py
arunkannawadi: The PSF coaddition, including dropping the use of `getAveragePosition`, is coming next in DM-43515.

arunkannawadi: Do we care about the ordering here when persisting. I can see if we try to compare if two files are identical without even opening them, but is that something we expect to support?

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -323,11 +345,30 @@ def run(self, inputWarps, skyInfo, **kwargs):
...
+                if True:
+                    ccd_table = warp.getInfo().getCoaddInputs().ccds.subsetContaining(cell_centers_sky[cellInfo.index])
+                    assert len(ccd_table) > 0, "No CCD from a warp found within a cell."
+                    assert len(ccd_table) == 1, "More than one CCD from a warp found within a cell."
+                    ccd_row = ccd_table[0]
+                else:
+                    for ccd_row in warp.getInfo().getCoaddInputs().ccds:
+                        if ccd_row.contains(cell_centers_sky[cellInfo.index]):
+                            break
+
```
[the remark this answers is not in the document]
arunkannawadi: > Make a Numpy array of cell centers in sky coordinates outside the loop over warps.

OK, we do this now but as a `dict` instead of an array.

> Compare those detector-coordinate positions to the detector bounding boxes.

We're going to have to do comparisons with polygons, since detectors aren't within rectangles anymore. I'll have to check that those comparisons are efficient as well.

But I'm going to create another ticket to improve the efficiency, since we are still very much limited by PSF evaluation per-cell.

python/lsst/drp/tasks/assemble_cell_coadd.py
TallJimbo: I'm not sure we can make it all the way to full support, but I think it's good practice to minimize unnecessary differences.

python/lsst/drp/tasks/assemble_cell_coadd.py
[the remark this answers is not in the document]
TallJimbo: > since detectors aren't within rectangles anymore.

That's why I suggested going all the way back to detector coordinates.  But doing this in polygons on the sky or in coadd coordinates may well be more efficient if it avoids the reverse-transform to check for bad extrapolation.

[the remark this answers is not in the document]
arunkannawadi: Ah, I misunderstood the coordinate system in your comment at first.

python/lsst/drp/tasks/assemble_cell_coadd.py
arunkannawadi: Alright, in the [other PR](https://github.com/lsst/cell_coadds/pull/31), I added a `__lt__` operator for `ObservationIdentifiers` and sort the inputs in the `__init__` method of `SingleCellCoadd`.

PR #48, opened 2024-04-05
DM-43515: Evaluate PSFs at the center of cells

arunkannawadi:


python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -113,14 +113,15 @@ class AssembleCellCoaddConfig(PipelineTaskConfig, pipelineConnections=AssembleCe
         inclusiveMin=True,
         inclusiveMax=False,
     )
-    # The following config options are specific to the CoaddPsf.
-    coadd_psf = ConfigField(
-        doc="Configuration for CoaddPsf",
-        dtype=CoaddPsfConfig,
+    psf_warper = ConfigField(
+        doc="Configuration for the warper that warps the PSFs. It must have the same configuration used to "
+        "warp the images.",
```
arunkannawadi, line 118: We will have to enforce this via a contract in the pipeline, since I don't think there's any other way to enforce consistency across tasks.

TallJimbo: I agree that's the best approach for now, but I think it's also a nudge in the direction of moving PSF warping down to `MakeDirectWarpTask` (by having that warp yet another image plane that would for now have to be yet another separate dataset type).  I don't think we should consider doing that on this ticket, but between configuration consistency, free parallelism over visits, and ruling out differences in how the WCS is interpolated during warping, I think that's the direction we want to go eventually.

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -340,30 +345,49 @@ def run(self, inputWarps, skyInfo, **kwargs):
...
+                    srcWcs=ccd_row.getWcs(),
+                    destBBox=psf_bbox_gc[cellInfo.index],
+                )
+
+                # There may be NaNs in the PSF image. Set them to 0.0
+                warped_psf_maskedImage.variance.array[np.isnan(warped_psf_maskedImage.image.array)] = 1.0
+                warped_psf_maskedImage.image.array[np.isnan(warped_psf_maskedImage.image.array)] = 0.0
+
+                psf_stacker = psf_gc[cellInfo.index]
+                psf_stacker.add_masked_image(warped_psf_maskedImage, weight=weight)
```
TallJimbo, line 389: Are we using a `MaskedImage` here due to limitations in `AccumulatorMeanStacker`?

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -340,30 +345,49 @@ def run(self, inputWarps, skyInfo, **kwargs):
...
+                warped_psf_maskedImage = self.psf_warper.warpImage(
+                    destWcs=skyInfo.wcs,
+                    srcImage=undistorted_psf_maskedImage,
+                    srcWcs=ccd_row.getWcs(),
+                    destBBox=psf_bbox_gc[cellInfo.index],
+                )
+
+                # There may be NaNs in the PSF image. Set them to 0.0
+                warped_psf_maskedImage.variance.array[np.isnan(warped_psf_maskedImage.image.array)] = 1.0
+                warped_psf_maskedImage.image.array[np.isnan(warped_psf_maskedImage.image.array)] = 0.0
```
TallJimbo, line 386: I think you can pass a custom fill value to `warpImage` instead of having to do this.

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -340,30 +345,49 @@ def run(self, inputWarps, skyInfo, **kwargs):
...
+                undistorted_psf_im = ccd_row.getPsf().computeImage(calexp_point)
+                # Adjust the bounding box of undistorted_psf_im to be centered
+                # at the closest integer location to the requested location.
+                undistorted_psf_im.setXY0(
+                    geom.Box2I.makeCenteredBox(calexp_point, undistorted_psf_im.getDimensions()).getMin()
+                )
+
+                # Convert the PSF image from Image to MaskedImage.
+                undistorted_psf_maskedImage = afwImage.MaskedImageD(image=undistorted_psf_im)
+                # TODO: In DM-43585, use the variance plane value from noise.
```
TallJimbo, line 374: I don't think it matters; the stacker had better not be using this variance for anything other than the output variance (we're passing it an explicit weight below), and we're throwing the output variance away.

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -340,30 +345,49 @@ def run(self, inputWarps, skyInfo, **kwargs):
...
+                stacker = gc[cellInfo.index]
+                stacker.add_masked_image(mi, weight=weight)
+
+                calexp_point = ccd_row.getWcs().skyToPixel(cell_centers_sky[cellInfo.index])
+                undistorted_psf_im = ccd_row.getPsf().computeImage(calexp_point)
+                # Adjust the bounding box of undistorted_psf_im to be centered
+                # at the closest integer location to the requested location.
+                undistorted_psf_im.setXY0(
+                    geom.Box2I.makeCenteredBox(calexp_point, undistorted_psf_im.getDimensions()).getMin()
+                )
```
TallJimbo, line 370: I don't think we want this `setXY0`; `Psf.computeImage` should already be setting xy0 such that the PSF image we've just rendered looks like a star at the right position, and that's what the `warpImage` call below expects.  I don't understand why it isn't causing trouble in your testing, though, unless it turns out to be a no-op or near no-op.

python/lsst/drp/tasks/assemble_cell_coadd.py
arunkannawadi: I have to double check, but we need that variance because its inverse is used as the weight.

python/lsst/drp/tasks/assemble_cell_coadd.py
arunkannawadi: We use `MaskedImage` mainly it warps differently from an `Image` (known issue, unknown cause). I could pick up the `image` plane and call `add_image` here, but given that the conversion is already made, I figured this keeps the code paths for images and PSFs least divergent.

python/lsst/drp/tasks/assemble_cell_coadd.py
arunkannawadi: I see that to be the natural progression, given that we've moved PSF evaluation from measurement to coaddition and the next natural step is to move it to `make_direct_warp.py`. We spoke about this, but my thought here has been to use `MultipleCellCoadd` (we might need to subclass it, more on that later) with no buffer between cells to be the data structure to hold warps. It has the capability to hold multiple noise image planes which are used by cell-based coadds as well. The PSF images can be evaluated at the cell centers at the time of warping.

The reason why we might need another class - or make a generic parent class to subclass from - is that we now have to allow for multiple detectors within each cell (we don't want to discard them just yet). This needs some minor changes to how we serialize metadata info.

But to reiterate again,
> differences in how the WCS is interpolated during warping

is not an issue because we evaluate the PSF image on the calexp and warp them here.

TallJimbo: > But to reiterate again,
>>    differences in how the WCS is interpolated during warping
>
> is not an issue because we evaluate the PSF image on the calexp and warp them here.

We don't evaluate the PSF image on the _full_ calexp; we evaluate it into a much smaller image in calexp coordinates.  While that would be sufficient if warping used the real WCS for every pixel, or if it evaluated the WCS at fixed points in PARENT coordinates (accounting for xy0) and interpolated between them.  But I suspect that it might actually evaluate the WCS at fixed points in LOCAL coordinates (i.e. not accounting for xy0) and interpolate those.  Just suspicion, though; if you know that it evaluates the PSF at fixed points in PARENT coordinates then this concern is indeed unfounded.

python/lsst/drp/tasks/assemble_cell_coadd.py
enourbakhsh: I recently tried something similar with `warpImage` in C++, thinking the last argument was for the fill value. But then I figured out it's actually for padding.  So, not sure if we can specify a fill value there unless I've misunderstood something.

arunkannawadi: What Erfan said. I _think_ the underlying interpolator probably has a `fill_value` argument, but `warpImage` does not. It'd be good to have this feature though.

python/lsst/drp/tasks/assemble_cell_coadd.py
arunkannawadi: It turned out to be a no-op indeed, since the bbox I am setting is the bbox that it already has.

python/lsst/drp/tasks/assemble_cell_coadd.py
arunkannawadi: Isn't this related to your comment about `xy0` below? The bounding box of the PSF image is evaluated in PARENT coordinates by default but I wonder if I'm missing your point. Happy to discuss this more tomorrow.

TallJimbo: This question doesn't need to hold up this branch anyway, and I think the answer is one that can be found in the depths of the warping code: we're making different assumptions about what it's doing, but the truth is out there.

TallJimbo (approved):

PR #49, opened 2024-04-11
DM-39841: Remove the use of doPsfMatch

arunkannawadi:


python/lsst/drp/tasks/assemble_coadd.py
```diff
@@ -302,6 +296,8 @@ def validate(self):
                 "statistic %s is not allowed. Please choose one of %s." % (self.statistic, stackableStats)
             )
 
+        super().validate()
```
erykoff, line 299: Is there a reason you moved this?  As far as I can tell, the pattern everywhere is for `super().validate()` to come first.

arunkannawadi: Hmm, I thought calling the parent class's `validate()` method in the end is preferred so it'll fail faster if it's going to fail.

PR #50, opened 2024-04-12
DM-43771: disable warping kernel caching and make sure we do PSF-evaluation caching.

TallJimbo:


python/lsst/drp/tasks/forcedPhotCoadd.py
```diff
@@ -320,6 +320,12 @@ def run(self, measCat, exposure, refCat, refWcs, exposureId=None):
                 Catalog of forced measurement results
                 (`lsst.afw.table.SourceCatalog`).
         """
+        # We want to cache repeated PSF evaluations at the same point coming
+        # from different measurement plugins.  We assume each algorithm tries
+        # to evaluate the PSF twice, which is more than enough since many don't
+        # evaluate it at all, and there's no *good* reason for any algorithm to
+        # evaluate it more than once.
+        exposure.psf.setCacheCapacity(2 * len(self.config.measurement.plugins.names))
```
arunkannawadi, line 328: Is the rationale for 2 here is to cache the resulting image from calling `computeImage` and `computeKernelImage`?

arunkannawadi (approved):

python/lsst/drp/tasks/forcedPhotCoadd.py
TallJimbo: No, those are already separate caches, so they effectively already have that factor of two.  I'm just being defensive against plugins doing more PSF evaluations than they probably should.

arunkannawadi: Ah, yes. Right!

PR #57, opened 2024-06-01
DM-44623: Use PSF-matched polygons from the data itself

arunkannawadi:


python/lsst/drp/tasks/assemble_coadd.py
```diff
@@ -379,7 +379,22 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
         else:
             warpRefList = inputData["inputWarps"]
 
-        inputs = self.prepareInputs(warpRefList)
+        # Although psfMatchedWarps are specifically required by only by a
+        # specific set of a subclass, and since runQuantum is not overridden,
+        # the selection and filtering must happen here on a conditional basis.
+        # Otherwise, the elements in the various lists will not line up.
```
TallJimbo, line 385: Might be worth extending this comment to make it clear that this does indeed suggest that  `AssembleCoadd` and `CompareWarpAssembleCoadd` need to be refactored; you could reference DM-38630 as a blocker, if not necessarily a TODO ticket (I'm definitely not suggesting it do that as part of this ticket!).

python/lsst/drp/tasks/assemble_coadd.py
```diff
@@ -495,8 +511,13 @@ def prepareInputs(self, refList):
         tempExpRefList = []
         weightList = []
         imageScalerList = []
+        outputPsfMatchedWarpRefList = []
+
+        if not psfMatchedWarpRefList:
+            psfMatchedWarpRefList = (None for _ in range(len(refList)))
```
TallJimbo: I think it's better to default to list comprehension rather than generator expression; lists just behave less surprisingly and are usually more performant when the thing you're iterating over isn't absolutely enormous.

python/lsst/drp/tasks/assemble_coadd.py
```diff
@@ -457,7 +473,7 @@ def _makeSupplementaryData(self, butlerQC, inputRefs, outputRefs):
     def makeSupplementaryDataGen3(self, butlerQC, inputRefs, outputRefs):
         return self._makeSupplementaryData(butlerQC, inputRefs, outputRefs)
 
-    def prepareInputs(self, refList):
+    def prepareInputs(self, refList, psfMatchedWarpRefList=None):
```
TallJimbo, line 476: Need to update the docstring as well.

python/lsst/drp/tasks/assemble_coadd.py
```diff
@@ -570,9 +595,11 @@ def prepareStats(self, mask=None):
     def run(
         self,
         skyInfo,
+        *,
         tempExpRefList,
         imageScalerList,
         weightList,
+        psfMatchedWarpRefList=None,
```
TallJimbo, line 602: Need to update docstring as well.

python/lsst/drp/tasks/dcr_assemble_coadd.py
```diff
@@ -498,7 +500,16 @@ def prepareDcrInputs(self, templateCoadd, warpRefList, weightList):
...
+    def run(
+        self,
+        skyInfo,
+        tempExpRefList,
+        imageScalerList,
+        weightList,
+        psfMatchedWarpRefList=None,
+        supplementaryData=None,
+        **kwargs
+    ):
```
TallJimbo, line 512: Need to update docstring to include psfMAtchedWarpRefList.

python/lsst/drp/tasks/dcr_assemble_coadd.py
```diff
@@ -527,7 +538,7 @@ def run(self, skyInfo, warpRefList, imageScalerList, weightList, supplementaryDa
         ----------
         skyInfo : `lsst.pipe.base.Struct`
             Patch geometry information, from getSkyInfo
-        warpRefList : `list` of `lsst.daf.butler.DeferredDatasetHandle`
+        tempExpRefList : `list` of `lsst.daf.butler.DeferredDatasetHandle`
```
TallJimbo, line 541: Why switch terminology here?  Generally I consider `tempExp` to be the historical name that we're moving away from, and `warp` the newer name we're moving towards.

TallJimbo (approved):

python/lsst/drp/tasks/assemble_coadd.py
arunkannawadi: Hmm, I thought I had it changed to list comprehension.

TallJimbo: Ah, I did review some of this commit-by-commit, so maybe I just commented on code you already changed before you were done.

python/lsst/drp/tasks/assemble_coadd.py
```diff
@@ -495,8 +511,13 @@ def prepareInputs(self, refList):
         tempExpRefList = []
         weightList = []
         imageScalerList = []
+        outputPsfMatchedWarpRefList = []
+
+        if not psfMatchedWarpRefList:
+            psfMatchedWarpRefList = [None for _ in range(len(refList))]
```
arunkannawadi, line 517: Yup, this is a list comprehension ultimately. I will also change it to just `[None] * len(refList)`.

PR #59, opened 2024-06-12
DM-44790: Deprecate doUsePsfMatchedPolygons

arunkannawadi:


python/lsst/drp/tasks/assemble_coadd.py
```diff
@@ -328,6 +336,9 @@ class AssembleCoaddTask(CoaddBaseTask, pipeBase.PipelineTask):
     ConfigClass = AssembleCoaddConfig
     _DefaultName = "assembleCoadd"
 
+    _doUsePsfMatchedPolygons: bool = False
+    """Use ValidPolygons from shrunk Psf-Matched Calexps?"""
+
```
arunkannawadi, line 341: A safer way would be to make this an immutable property. However, I think marking this with a leading underscore is indication enough that this shouldn't be meddled with, while allowing for the possibility of toggling it for experimental reasons.

python/lsst/drp/tasks/assemble_coadd.py
```diff
@@ -1415,6 +1428,9 @@ class CompareWarpAssembleCoaddTask(AssembleCoaddTask):
     ConfigClass = CompareWarpAssembleCoaddConfig
     _DefaultName = "compareWarpAssembleCoadd"
 
+    _doUsePsfMatchedPolygons: bool = True
```
taranu, line 1431: Why are you re-defining the attribute instead of just setting `self._doUsePsfMatchedPolygons = True` in `__init__`?

python/lsst/drp/tasks/assemble_coadd.py
```diff
@@ -138,10 +140,14 @@ def __init__(self, *, config=None):
 class AssembleCoaddConfig(
     CoaddBaseTask.ConfigClass, pipeBase.PipelineTaskConfig, pipelineConnections=AssembleCoaddConnections
 ):
-    warpType = pexConfig.Field(
+    warpType = pexConfig.ChoiceField(
```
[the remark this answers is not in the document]
arunkannawadi: I don't think this breaks backwards compatibility at all. Trying to read in old config files would just work fine, as long as the values they have are one of the allowed values. These are the really two ones we support, and if the value was anything else, only then would it break (and it would be real error anyway). One way to see it is that we now have an extra constraint in the `validate` method, except that the check would happen much earlier.

python/lsst/drp/tasks/assemble_coadd.py
```diff
@@ -1016,12 +1030,12 @@ def applyAltMaskPlanes(self, mask, altMaskSpans):
...
             Updated mask.
         """
-        if self.config.doUsePsfMatchedPolygons:
+        if self._doUsePsfMatchedPolygons:
             if ("NO_DATA" in altMaskSpans) and ("NO_DATA" in self.config.badMaskPlanes):
                 # Clear away any other masks outside the validPolygons. These
                 # pixels are no longer contributing to inexact PSFs, and will
                 # still be rejected because of NO_DATA.
-                # self.config.doUsePsfMatchedPolygons should be True only in
+                # self._doUsePsfMatchedPolygons should be True only in
```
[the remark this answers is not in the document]
arunkannawadi: Well, the `CompareWarpAssembleCoaddTask` and its child classes, which is really only `DcrAssembleCoaddTask` and that one is really not been used anywhere right now. So yes, it is true.

python/lsst/drp/tasks/assemble_coadd.py
arunkannawadi: I'd like to be able to get this class attribute without necessarily having to create an instance - i.e., I'd like `CompareWarpAssembleCoaddTask._doUsePsfMatchedPolygons` to already return True. The only minor downside is having to define the docstring twice - which is mostly to please the editor that Sphinx. If we go the read-only class property route, I know the docstring inheritance works for sure.

python/lsst/drp/tasks/assemble_coadd.py
arunkannawadi: I shall. But would you have felt the same way if I had defined this is as

```python
@property
@classmethod
def _doUsePsfMatchedPolygons(cls):
  return False
```

and the same with `True` in `CompareWarpAssembleCoaddTask` ?

taranu: Well in that case you couldn't turn it on at all, so I guess any such note would only be useful if you wanted to advertise that it can be turned on in a derived class if... (whatever the conditions are).

python/lsst/drp/tasks/assemble_coadd.py
taranu: I see. It seems like you want a truly static and immutable default (which a classmethod property is good for) but I'm not so clear on whether you want a variable class/instance attribute. Does it make sense to turn this off for `CompareWarpAssembleCoaddTask`? If so then I think you want both. Likewise for whether it ought to be a class or instance attribute, I'd say the latter is generally safer even if we don't tend to make two instances of the same pipelineTask very often.

python/lsst/drp/tasks/assemble_coadd.py
arunkannawadi: Oh, it's certainly mutable :-) You _could_ turn it on by setting
```python
AssembleCoaddTask._doUsePsfMatchedPolygons = property(lambda x: True)
```

I just don't expect anyone to turn it on and at least much less than anyone wanting to set `_doUsePsfMatchedPolygons`.

But this highlights the reason to move this away from `config`, which is that it should really be considered as a core part of the algorithm itself. The fact that this variable is needed is because of shared `run` method.

python/lsst/drp/tasks/assemble_coadd.py
arunkannawadi: I am considering this to be on par with the `_DefaultName` just above, which should just be treated as a constant for this class. Turning it on/off in either task would run fine - changing from the default values is just not the correct thing to do given the algorithm(s).

python/lsst/drp/tasks/assemble_coadd.py
taranu: I don't think that's actually mutating but reassigning, and hopefully we haven't merged any awful code like that without good reason.

taranu: Then just make it a classmethod only, it's not that important tbat it be a property or attr.

arunkannawadi: No, I think we agreed that leading underscore suffices.

arunkannawadi:
I've added the criterion when `_doUsePsfMatchedPolygons` should be turned on. I didn't mention that child class explicitly, because a parent class known nothing about its children class (converse is not true).

All I want to do in this ticket is to remove it from the `Config`. Anyone who knows what they are doing can toggle this whenever. The code would run till the end without crashing in either mode. So I'm not inclined to make it a method.

python/lsst/drp/tasks/assemble_coadd.py
taranu: I thought you wanted it to be immutable, in which case it doesn't really matter if there's a leading underscore or not.

If you still want it to be mutable, fine, but if it's going to do the wrong thing when changed from the default then I'm not really sure what the benefit of that is.

arunkannawadi: I just didn't want a naive user to go about changing the value, I don't care as much about the attribute itself being immutable. Being a `Config` field is an open invitation to change the value, which is now avoided. Leading underscore suggests that it _is possible_ to change the value at your own risk, but you _should not_ . And by you, I mean a user of this task, not a developer modifying this task.

PR #60, opened 2024-06-28
DM-45028:  Resolve ID mismatch error in assembleCoadd

arunkannawadi:


python/lsst/drp/tasks/assemble_coadd.py
```diff
@@ -1105,25 +1105,40 @@ def _setValidPolygons(self, warpList, psfMatchedWarpRefList):
...
-                assert ccdTable[idx].id == psfMatchedCcdTable[idx].id, "ID mismatch"
-                if not psfMatchedCcdTable[idx].validPolygon:
+
+            # In some (literal) edge cases, a small part of the CCD may be
+            # present in directWarp that gets excluded in psfMatchedWarp. It is
+            # okay to leave validPolygon for those CCDs empty. However, the
+            # converse is not expected, and raise an error if that is the case.
+            if not set(psfMatchedCcdTable["id"]).issubset(ccdTable["id"]):
+                visit = psfMatchedWarpRef.dataId["visit"]
+                raise ValueError(f"PSF-matched warp for has additional CCDs for {visit=}")
```
leeskelvin, line 1122: raise typo.

python/lsst/drp/tasks/assemble_coadd.py
```diff
@@ -1105,25 +1105,40 @@ def _setValidPolygons(self, warpList, psfMatchedWarpRefList):
...
-            psfMatchedCcdTable.sort()
-            ccdTable.sort()
-            for idx in range(len(ccdTable)):
-                assert ccdTable[idx].id == psfMatchedCcdTable[idx].id, "ID mismatch"
-                if not psfMatchedCcdTable[idx].validPolygon:
+
+            # In some (literal) edge cases, a small part of the CCD may be
+            # present in directWarp that gets excluded in psfMatchedWarp. It is
+            # okay to leave validPolygon for those CCDs empty. However, the
+            # converse is not expected, and raise an error if that is the case.
```
leeskelvin, line 1119: "raise" -> "will raise" or "raises"

(I normally wouldn't comment about something so small, but since you're going to touch this file again, I thought I would).

leeskelvin (approved):

python/lsst/drp/tasks/assemble_coadd.py
arunkannawadi: I had meant `raise` as an imperative to the developer :-D ... and I don't think I'd follow the logic of what is going down here if I revisited a year from now, so I figured I'll add a comment.

leeskelvin: I think the "and" makes this a bit of a garden path sentence. How about this? 

> However, the converse is not expected; raise an error if that is the case.

arunkannawadi: I just changed this to be in passive.

PR #61, opened 2024-07-12
DM-45224: Deep copy the visit summary before modifying it

timj:
Without this if the tests run in a different order some of them fail because the WCS was deleted.


cmsaunders (commented):
LGTM. Thanks for catching this--the way it is currently done was definitely an oversight on my part.

PR #64, opened 2024-07-29
DM-45426: Account for warps from make_direct_warp are bigger

arunkannawadi:


python/lsst/drp/tasks/assemble_coadd.py
```diff
@@ -392,6 +392,9 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
             skyMap, tractId=outputDataId["tract"], patchId=outputDataId["patch"]
         )
 
+        # Attach the bounding box of the coadd this task is producing.
+        self.coadd_bbox = inputData["skyInfo"].bbox
```
TallJimbo, line 396: I'm not a fan of putting per-quantum state on the task instance.  Better to pass this around as another argument if it's needed in multiple methods.

TallJimbo (approved):

PR #65, opened 2024-08-02
DM-39168: Add utility to compute apparent motion

cmsaunders:


python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -57,6 +58,66 @@
...
+        Dec shift in degrees.
+    """
+    ra_rad = (data["ra"].values * u.deg).to(u.rad)
+    dec_rad = (data["dec"].values * u.deg).to(u.rad)
+
+    dt = (astropy.time.Time(data["MJD"].values, format="mjd") - astropy.time.Time(refEpoch, format="mjd")).to(
+        u.yr
+    )
+    pmRA_deg = (data["pmRA"].values * u.mas).to(u.deg) / u.yr
+    pmDec_deg = (data["pmDec"].values * u.mas).to(u.deg) / u.yr
```
parejkoj, line 92: This is another set of unit assumptions that should be documented in the `data` docstring (that pm values should be in mas).

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -57,6 +58,66 @@
...
+    properMotionRA = pmRA_deg * dt
+    properMotionDec = pmDec_deg * dt
+
+    obsTimes = astropy.time.Time(data["MJD"], format="mjd")
```
parejkoj, line 96: Why not assing this above, and then use it in `dt`? Otherwise, you're making this time object twice. Or does it make more sense to have the observed epoch be a separate input argument, as a Time object already?

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -57,6 +58,66 @@
...
+    sun = astropy.coordinates.get_body("sun", time=obsTimes)
+    frame = astropy.coordinates.GeocentricTrueEcliptic(equinox=obsTimes)
+    sunLongitudes = sun.transform_to(frame).lon.radian
+
+    # These equations for parallax come from Equations 5.2 in Van de Kamp's
+    # book Stellar Paths. They differ from the parallax calculated in gbdes by
+    # ~0.01 mas, which is acceptable for QA and plotting purposes.
+    parallaxFactorRA = np.cos(wcsfit.EclipticInclination) * np.cos(ra_rad) * np.sin(sunLongitudes) - np.sin(
```
parejkoj, line 104: `wcsfit` is a gbdes thing, no? Could you get this constant from astropy instead?

tests/test_gbdesAstrometricFit.py
```diff
@@ -1058,6 +1061,114 @@ def test_inputCameraModel(self):
...
+
+        cls.mjds = np.random.rand(10) * 20 + 57000
+
+    def testProperMotion(self):
+        """Calculate the change in position due to proper motion for a given
+        time change, and compare the results against astropy.
+        """
+
+        parallaxes = np.zeros(len(self.RAs))
+        refEpoch = 57100
```
parejkoj, line 1090: I see you use this in the mjds you assign above. You should should make it an instance variable, because otherwise it looks like a magic constant there.

tests/test_gbdesAstrometricFit.py
```diff
@@ -1058,6 +1061,114 @@ def test_inputCameraModel(self):
...
 
 
+class TestApparentMotion(lsst.utils.tests.TestCase):
+    """This class simulates an array of objects with random proper motions and
+    parallaxes and compares the results of
+    `lsst.drp.tasks.gbdesAstrometricFit.calculate_apparent_motion` against the
+    results calculated by astropy."""
+
+    @classmethod
+    def setUpClass(cls):
```
parejkoj, line 1071: I don't see any real reason to use `setUpClass` here, why not just `setUp`? The main reason for `setUpClass` is for things that take a long time.

tests/test_gbdesAstrometricFit.py
```diff
@@ -1058,6 +1061,114 @@ def test_inputCameraModel(self):
...
+        cls.parallaxes = abs(np.random.random(10) * 5)
+
+        cls.mjds = np.random.rand(10) * 20 + 57000
+
+    def testProperMotion(self):
+        """Calculate the change in position due to proper motion for a given
+        time change, and compare the results against astropy.
+        """
+
+        parallaxes = np.zeros(len(self.RAs))
```
parejkoj, line 1089: I'd add a comment here that this test is not of the parallax correction, so it's more obvious why this is zero.
```suggestion
        parallaxes = np.zeros_like(self.RAs)
```

tests/test_gbdesAstrometricFit.py
```diff
@@ -1058,6 +1061,114 @@ def test_inputCameraModel(self):
...
+    parallaxes and compares the results of
+    `lsst.drp.tasks.gbdesAstrometricFit.calculate_apparent_motion` against the
+    results calculated by astropy."""
+
+    @classmethod
+    def setUpClass(cls):
+        np.random.seed(12345)
+
+        cls.RAs = np.random.rand(10) + 150.0
+        cls.Decs = np.random.rand(10) + 2.5
```
parejkoj, line 1075: No first letter caps on things that aren't classes. `self.ras` and `self.decs`. Or even not-plural `self.ra`, since it doesn't really matter that they are arrays (similarly for all the other variables here).

tests/test_gbdesAstrometricFit.py
```diff
@@ -1058,6 +1061,114 @@ def test_inputCameraModel(self):
...
+
+        cls.parallaxes = abs(np.random.random(10) * 5)
+
+        cls.mjds = np.random.rand(10) * 20 + 57000
+
+    def testProperMotion(self):
+        """Calculate the change in position due to proper motion for a given
+        time change, and compare the results against astropy.
+        """
+
```
parejkoj, line 1088: Unnecessary line break.

tests/test_gbdesAstrometricFit.py
```diff
@@ -1058,6 +1061,114 @@ def test_inputCameraModel(self):
...
+    @classmethod
+    def setUpClass(cls):
+        np.random.seed(12345)
+
+        cls.RAs = np.random.rand(10) + 150.0
+        cls.Decs = np.random.rand(10) + 2.5
+
+        cls.pmRAs = np.random.random(10) * 10
+        cls.pmDecs = np.random.random(10) * 10
```
parejkoj, line 1078: You've got a mix of camelCase and snake_case in this file (e.g. `ra_correction` below). We're moving to snake_case in general, so it might be worth doing that here.
```suggestion
        cls.pm_ra = np.random.random(10) * 10
        cls.pm_dec = np.random.random(10) * 10
```

tests/test_gbdesAstrometricFit.py
```diff
@@ -1058,6 +1061,114 @@ def test_inputCameraModel(self):
...
+                "MJD": self.mjds,
+            }
+        )
+        ra_correction, dec_correction = calculate_apparent_motion(data, refEpoch)
+
+        pmRA_cosDec = self.pmRAs * np.cos((self.Decs * u.degree).to(u.radian))
+        coords = SkyCoord(
+            ra=self.RAs * u.degree,
+            dec=self.Decs * u.degree,
+            pm_ra_cosdec=pmRA_cosDec * u.mas / u.yr,
```
parejkoj, line 1108: Why do you multiply by `u.mas` here, it's already a Quantity above, no? Why not do this?

```suggestion
            pm_ra_cosdec=self.pm_ra * np.cos((self.decs * u.degree) / u.yr
```

tests/test_gbdesAstrometricFit.py
```diff
@@ -1058,6 +1061,114 @@ def test_inputCameraModel(self):
...
+            pm_dec=pms * u.mas / u.yr,
+            distance=Distance(parallax=self.parallaxes * u.mas),
+            obstime=Time(refEpoch, format="mjd"),
+        )
+        refCoords = SkyCoord(
+            ra=self.RAs * u.degree,
+            dec=self.Decs * u.degree,
+            pm_ra_cosdec=pms * u.mas / u.yr,
+            pm_dec=pms * u.mas / u.yr,
+            distance=1e10 * Distance(parallax=self.parallaxes * u.mas),
```
parejkoj, line 1157: Why 1e10?

parejkoj (changes requested):
Just to check: why can't you use astropy to do the PM+parallax correction directly? I see you're doing just that in the tests; would it make sense to just make this function a wrapper around `astropy.SkyCoord` conversions?

I'll trust you on the parallax math itself: I don't have the reference to hand.

cmsaunders:
It's not possible to use the `astropy.SkyCoord` conversions here for just parallax because they include other apparent motions like annual aberration. (See this github issue discussing the topic: https://github.com/astropy/astropy/issues/9140.) Annual aberration is one of the effects that is accounted for by the overall per-visit part of the astrometric model, so we do not want it included here, where we are looking for only the correction to the position of a source relative to the other sources in the field.
There is a sort of hack in `astropy.SkyCoord`, where you can get motion due to parallax by itself by comparing a nearby and distant star. I don't think it's sufficient for this to be the actual way we do the correction, but I use it in the test because I don't have another good way to confirm the parallax calculation.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -57,6 +58,66 @@
 ]
 
 
+def calculate_apparent_motion(data, refEpoch):
+    """Calculate shift from reference epoch to the apparent observed position
+    at another date.
+
+    This function calculates the shift due to proper motion combined with the
+    apparent motion due to parallax. This is not used in the
+    `GbdesAstrometricFitTask` or related child tasks, but is useful for
```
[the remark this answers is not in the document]
cmsaunders: I don't mind doing that, except that this is specifically geared to match what is in `gbdes`. It is calculating the motion relative to other stars in the field, not the actual apparent motion, which would also include annual aberration. I am also using things like the ecliptic angle hard-coded in `gbdes` rather than one of the astropy options to better match what is coming out of the fitter.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -57,6 +58,66 @@
...
+        source.
+    refEpoch : `float`
+        Epoch of the reference position.
+
+    Returns
+    -------
+    apparentMotionRA : `np.ndarray` [`astropy.units.Quantity`]
+        RA shift in degrees.
+    apparentMotionDec : `np.ndarray` [`astropy.units.Quantity`]
+        Dec shift in degrees.
```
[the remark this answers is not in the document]
cmsaunders: We store coord_ra and coord_dec as degrees, so I think it makes sense for the correction that will be applied to them to be in degrees. I think it's helpful to say in the docs that the Quantity is in degrees.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -57,6 +58,66 @@
...
+    This function calculates the shift due to proper motion combined with the
+    apparent motion due to parallax. This is not used in the
+    `GbdesAstrometricFitTask` or related child tasks, but is useful for
+    assessing results.
+
+    Parameters
+    ----------
+    data : `pd.DataFrame`
+        Table containing position, proper motion, parallax, and epoch for each
+        source.
```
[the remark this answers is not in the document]
cmsaunders: I added the required names for the fields. We don't have any data products that this could be directly run on. In `analysis_tools`, I am constructing the catalog that will be used.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -57,6 +58,66 @@
...
+        RA shift in degrees.
+    apparentMotionDec : `np.ndarray` [`astropy.units.Quantity`]
+        Dec shift in degrees.
+    """
+    ra_rad = (data["ra"].values * u.deg).to(u.rad)
+    dec_rad = (data["dec"].values * u.deg).to(u.rad)
+
+    dt = (astropy.time.Time(data["MJD"].values, format="mjd") - astropy.time.Time(refEpoch, format="mjd")).to(
+        u.yr
+    )
```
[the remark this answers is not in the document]
cmsaunders: Yeah, I don't love it, but `black` is enforced and run automatically on this repo.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: I think I just missed that I was using the identical conversion twice.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: As commented elsewhere, I am trying to replicate `gbdes` as closely as possible, and `astropy` doesn't give the same value.

tests/test_gbdesAstrometricFit.py
```diff
@@ -1058,6 +1061,114 @@ def test_inputCameraModel(self):
...
+        )
+        ra_correction, dec_correction = calculate_apparent_motion(data, refEpoch)
+
+        pmRA_cosDec = self.pmRAs * np.cos((self.Decs * u.degree).to(u.radian))
+        coords = SkyCoord(
+            ra=self.RAs * u.degree,
+            dec=self.Decs * u.degree,
+            pm_ra_cosdec=pmRA_cosDec * u.mas / u.yr,
+            pm_dec=self.pmDecs * u.mas / u.yr,
+            obstime=Time(refEpoch, format="mjd"),
```
timj, line 1110: General point -- be careful with timescales. If you know you are using tai everywhere you might want to specify that. If you know it's UTC somewhere, then say so. All your `Time` objects are defaulting to UTC and that might not be what you mean given that all our times in FITS metadata and elsewhere are TAI.

tests/test_gbdesAstrometricFit.py
cmsaunders: Is there a downside to `setUpClass`? It just seems preferable to not rerun the setup if it's not necessary.

tests/test_gbdesAstrometricFit.py
cmsaunders: `np.zeros_like` is much slower that `np.zeros`. Not a big deal here because `self.RAs` is not big, but I usually avoid zeros_like and ones_like.

tests/test_gbdesAstrometricFit.py
cmsaunders: I guess I don't need to convert to radians at line 1104, but after you do `np.cos((self.decs * u.degree)`, you have a dimensionless quantity, so it still needs to have the `u.mas`.

tests/test_gbdesAstrometricFit.py
cmsaunders: I've gone through and made the capitalization consistent with the rest of the `gbdesAstrometricFit` module, which is snake_case for function names and camelCase otherwise.

tests/test_gbdesAstrometricFit.py
```diff
@@ -32,13 +32,16 @@
 import pandas as pd
 import wcsfit
 import yaml
+from astropy.coordinates import Distance, SkyCoord
+from astropy.time import Time
```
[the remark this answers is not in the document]
cmsaunders: Reasonable, especially since that is how it's done already in `gbdesAstrometricFit.py`.

tests/test_gbdesAstrometricFit.py
```diff
@@ -1058,6 +1061,114 @@ def test_inputCameraModel(self):
...
+            pm_ra_cosdec=pms * u.mas / u.yr,
+            pm_dec=pms * u.mas / u.yr,
+            distance=Distance(parallax=self.parallaxes * u.mas),
+            obstime=Time(refEpoch, format="mjd"),
+        )
+        refCoords = SkyCoord(
+            ra=self.RAs * u.degree,
+            dec=self.Decs * u.degree,
+            pm_ra_cosdec=pms * u.mas / u.yr,
+            pm_dec=pms * u.mas / u.yr,
```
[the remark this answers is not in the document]
cmsaunders: It's required if you want to use `apply_space_motion`.

tests/test_gbdesAstrometricFit.py
```diff
@@ -1058,6 +1061,114 @@ def test_inputCameraModel(self):
...
+
+        newCoords = coords.apply_space_motion(new_obstime=Time(self.mjds, format="mjd")).transform_to("gcrs")
+        refNewCoords = refCoords.apply_space_motion(new_obstime=Time(self.mjds, format="mjd")).transform_to(
+            "gcrs"
+        )
+        astropy_dRA = newCoords.ra.degree - refNewCoords.ra.degree
+        astropy_dDec = newCoords.dec.degree - refNewCoords.dec.degree
+
+        np.testing.assert_allclose(astropy_dRA, ra_correction.value, rtol=1e-1)
+        np.testing.assert_allclose(astropy_dDec, dec_correction.value, rtol=1e-1)
```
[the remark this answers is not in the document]
cmsaunders: Yeah, this is why I asked you to review the ticket! They're also both different from the value from `gbdes`, which is done via a change of coordinate systems. If I recall correctly, if you start digging into the astropy calculation, it takes you into a maze of ERFA code, and I never figured out exactly how the parallax part was being calculated. Maybe @timj has a better idea for how to approach this?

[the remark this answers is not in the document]
cmsaunders: The upshot is that the actual differences are all about -0.012 mas, which may come down to the choice of ecliptic angle between the function and astropy, and I have decided this is acceptable in this context.

tests/test_gbdesAstrometricFit.py
cmsaunders: The point is to compare an object with a much more distant object at the same RA/Dec because parallax will only affect the position of the nearby object. This is described in the docstring, but I'll add a comment here.

tests/test_gbdesAstrometricFit.py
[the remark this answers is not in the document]
timj: I am still worrying that your Times are all in UTC and would like confirmation that they really should be in UTC and not TAI.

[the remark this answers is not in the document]
cmsaunders: I'm still replying to comments and haven't pushed any changes yet. The times come from the `VisitTable` and should be TAI. This is endemic in `gbdesAstrometricFit`.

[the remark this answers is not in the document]
timj: Ok. So every astropy.time.Time you create will need `scale="tai"` added to it.

python/lsst/drp/tasks/gbdesAstrometricFit.py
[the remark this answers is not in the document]
cmsaunders: I changed the variable name and changed it to an astropy Table so that I don't have to make any units assumptions.

tests/test_gbdesAstrometricFit.py
parejkoj: `setUpClass` can cause confusion if a test modifies (often inadvertently!) a variable that was set, such that subsequent tests use the wrong value. When running in a multiprocessing environment (as we do with our pytest setup), `setUpClass` will run once per process, which could mean once per test. Unless you know that your setup is really slow, it's not worth the potential confusion of class variables potentially being modified in a test.

tests/test_gbdesAstrometricFit.py
parejkoj: Oh, huh, I didn't know that there was a speed difference.

tests/test_gbdesAstrometricFit.py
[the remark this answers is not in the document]
parejkoj: Thanks, changing to `assertFloatsAlmostEqual` should make the exact comparison more obvious. And Tim's right about requiring `scale` in the astropy time objects.

tests/test_gbdesAstrometricFit.py
```diff
@@ -1136,6 +1140,116 @@ def test_useColor(self):
...
+            distance=1e10 * Distance(parallax=self.parallaxes),
+            obstime=self.refEpoch,
+        )
+
+        newCoords = coords.apply_space_motion(new_obstime=self.mjds).transform_to("gcrs")
+        refNewCoords = refCoords.apply_space_motion(new_obstime=self.mjds).transform_to("gcrs")
+        astropyDRA = newCoords.ra.degree - refNewCoords.ra.degree
+        astropyDDec = newCoords.dec.degree - refNewCoords.dec.degree
+
+        self.assertFloatsAlmostEqual(astropyDRA, raCorrection.value, rtol=1e-1)
```
parejkoj, line 1249: As I noted above, you should do the comparison in `mas` and probably with `atol`, so we can see how accurate it is in "real" units. It's not obvious to me what a 10% accuracy means here, but it doesn't feel like it's very good?

parejkoj (commented):
A few more comments, particularly about the tests.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -57,6 +58,66 @@
...
+    `GbdesAstrometricFitTask` or related child tasks, but is useful for
+    assessing results.
+
+    Parameters
+    ----------
+    data : `pd.DataFrame`
+        Table containing position, proper motion, parallax, and epoch for each
+        source.
+    refEpoch : `float`
+        Epoch of the reference position.
```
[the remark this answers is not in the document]
cmsaunders: Forgot to comment before--this is an `astropy.time.Time` object now.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: Another where I forgot to comment--these columns now come from an astropy table now and have units, so there is no assumption made.

tests/test_gbdesAstrometricFit.py
cmsaunders: Well, the actual accuracy is about 1.1% at worst, but I rounded up rather than have the value be too tuned. I guess I could do 2e-2. For the atol, I'm not sure that it's preferable here, since you don't know what the values of `astropyDRA` and `raCorrection` are here. I'm going to just add a comment on the scale of the errors here, which correspond to 0.04 mas at most.

tests/test_gbdesAstrometricFit.py
```diff
@@ -1136,6 +1140,116 @@ def test_useColor(self):
...
+            pm_ra_cosdec=pmRACosDec,
+            pm_dec=self.pmDecs,
+            obstime=self.refEpoch,
+        )
+
+        newCoords = coords.apply_space_motion(new_obstime=self.mjds)
+        astropyDRA = newCoords.ra.degree - coords.ra.degree
+        astropyDDec = newCoords.dec.degree - coords.dec.degree
+
+        self.assertFloatsAlmostEqual(astropyDRA, raCorrection.value, rtol=1e-6)
```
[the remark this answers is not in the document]
cmsaunders: This is the relative difference though, so 1e-6 is unitless. I added a comment that the proper motions are on the scale of 0-3 mas/yr and the differences are on the scale of 1e-8 mas.

tests/test_gbdesAstrometricFit.py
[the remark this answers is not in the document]
cmsaunders: Right, I already made the `scale` change.

tests/test_gbdesAstrometricFit.py
parejkoj: Ah, thank you for that clarification. That's not as bad as I'd thought.

parejkoj (approved):

PR #69, opened 2024-08-21
DM-44352: Add subtask to build camera from astrometric model

cmsaunders:


python/lsst/drp/tasks/build_camera.py
```diff
@@ -0,0 +1,778 @@
...
+                        deviceParams.append(params)
+        deviceArray = np.vstack(deviceParams)
+
+        for element in mapTemplate["EXPOSURE"]["Elements"]:
+            for visit in visitList:
+                visitTemplate = element.replace("EXPOSURE", str(visit))
+                for k, params in mapParams.items():
+                    if re.fullmatch(visitTemplate, k):
+                        visitParams.append(params)
+        identityVisitParams = np.array([0, 1, 0, 0, 0, 1])
```
PFLeget, line 315: For my personal understanding why this is like this?

PFLeget (approved):
It looks good to me. I have the feeling in build_camera.py, writing minimization with JAX, it can make the code shorter, but this is details. As long as this is doing the job, it is good to me.

python/lsst/drp/tasks/build_camera.py
cmsaunders: This is an array of coefficients that correspond to the `c` terms in `x' = c_{0} + c_{1} x + c_{2} y` and `y' = c_{3} + c_{4} x + c_{5} y`, meaning that this transform will give `x' = x` and `y' = y`.
In `gbdes`, in order to break degeneracies, one of the visits gets picked to have an Identity map for the per-visit, so I have to add that in here, along with the map parameters that are actually fit.

PR #70, opened 2024-08-26
DM-44764: Write out artifact mask

arunkannawadi:


python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -80,8 +89,18 @@ class AssembleCellCoaddConnections(
...
+
+        if config is None:
+            return
+
+        if not config.do_apply_artifact_mask:
+            self.inputs.remove("artifactMasks")
+
 
 class AssembleCellCoaddConfig(PipelineTaskConfig, pipelineConnections=AssembleCellCoaddConnections):
+    do_apply_artifact_mask = Field[bool](doc="Apply artifact masks to the input warps?", default=True)
```
arunkannawadi, line 103: I introduced this as a way to remove `artifactMask` as an input to making cell-based coadds on simulated data. However, this had a surprising (but correct) behavior - if excluded some quanta that didn't have PSF-matched warps and hence artifactMask. This is potentially very confusing to those who don't understand the middleware. I was going to work some solution out in `adjustQuantum` by selecting only those quanta with PSF-matched warps. But for the simulations, those are not bound to be present either. So I am inclined to remove this option together and expect a trivial all-zero `artifactMask` instead.

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -66,6 +66,15 @@ class AssembleCellCoaddConnections(
         multiple=True,
     )
 
+    artifactMasks = Input(
+        doc="Artifact masks to be applied to the input warps",
+        name="artifactMask",
```
arunkannawadi, line 71: I've kept the name as just `artifactMask` after our discussions.

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -313,11 +332,21 @@ def run(self, inputWarps, skyInfo, **kwargs):
...
         # it completely overlaps.
-        for warpRef in inputWarps:
+        for warpRef, artifactMaskRef in zip(inputWarps, artifactMasks):
             warp = warpRef.get()
 
+            if artifactMaskRef is not None:
+                # Apply the artifact mask to the warp.
+                artifactMask = artifactMaskRef.get()
+                assert artifactMask.getMaskPlaneDict() == warp.mask.getMaskPlaneDict()
+                assert not np.any(warp.mask.array & artifactMask.array)
```
arunkannawadi, line 346: Do these statements cover the guards that you mentioned during our discussions. These should all be unnecessary given the global nature of our `Mask`s.

python/lsst/drp/tasks/assemble_cell_coadd.py
TallJimbo: I don't find it confusing, but if it doesn't solve the problem you wanted it to solve (or any other problem) then I agree it's dead weight.

tests/test_assemble_coadd.py
```diff
@@ -334,6 +336,52 @@ def testInputMap(self):
...
+        dataRefList = []
+        for dataRef, artifactMask in zip(self.dataRefList, compareWarpResult.altMaskList):
+            dataSet = dataRef.get()
+            compareWarpTask.applyAltMaskPlanes(dataSet.mask, artifactMask)
+            # Repackage the dataset into a DataRef.
+            dataRefList.append(
+                pipeBase.InMemoryDatasetHandle(
+                    dataSet, storageClass=dataRef.storageClass, copy=True, dataId=dataRef.dataId
+                )
+            )
```
TallJimbo, line 363: I'd recommend replacing `dataRef` here with `handle` or something, because `DataRef` is a type that no longer exists (it's from the Gen2 butler).  Might even be worth doing that in the task itself (as a separate commit).

TallJimbo (approved):

python/lsst/drp/tasks/assemble_cell_coadd.py
arunkannawadi: Since you reviewed, I realized I have to explicitly add `CLIPPED` and `REJECTED` mask planes again, since they are not native to `warp`. Do you still think these asserts are unnecessary?

tests/test_assemble_coadd.py
arunkannawadi: You mean change the variable name to say `handle` instead of `ref`, right?

python/lsst/drp/tasks/assemble_cell_coadd.py
TallJimbo: No, not really.  My thinking was:
 - it'd be more efficient to assert that `getPlaneBitMask(plane)` returns the same thing for all planes of interest;
 - that's a static method so it can't possibly fail;
 - if it's possible for the assert on `getPlaneMaskDict` to fail, then `getPlaneBitMask` would broken;
 - the second assert is possibly meaningful but unlikely to matter, and fairly expensive for something in that category.

tests/test_assemble_coadd.py
TallJimbo: Yes, exactly.

python/lsst/drp/tasks/assemble_cell_coadd.py
TallJimbo: `artifact_mask` is fine.  I think I used `warp_mask` as an indication of the coordinate system the mask was in, so maybe `warp_artifact_mask` or even `compare_warp_mask` may be worth considering.

I'm glad John and others have paid more attention to the DMTN-172 conventions than I have (at least in code reviews :slightly_smiling_face:); I lost steam on actually getting all of that RFC-approved but it's better to adhere to informal naming conventions than none.

tests/test_assemble_coadd.py
arunkannawadi: The term `ref` is baked in quite strongly in the main task, sometimes as keyword-only arguments. It would also result in some temporary mismatch in the convention if we don't change the name in some tasks in `pipe_tasks`. So for the moment, I've fixed this in tests and in docstrings. I'm afraid a complete change would have to happen later.

python/lsst/drp/tasks/assemble_cell_coadd.py
arunkannawadi: `compare_warp_artifact_mask`. No one said naming is easy.

PR #73, opened 2024-09-17
DM-43077: Convert all tasks to use CalibrateImageTask outputs

parejkoj:


python/lsst/drp/tasks/forcedPhotCoadd.py
```diff
@@ -159,14 +159,15 @@ def setDefaults(self):
...
+            # TODO DM-44658 and DM-45980: don't have STREAK propagated yet.
+            # "STREAK",
         ]
         self.measurement.plugins["base_PixelFlags"].masksFpCenter = [
             "CLIPPED",
             "SENSOR_EDGE",
             "REJECTED",
             "INEXACT_PSF",
-            "STREAK",
+            # "STREAK",
```
leeskelvin, line 170: Although only a few lines away from the prior `TODO`, this probably also needs a `TODO` inline comment to ensure that it isn't missed when the time comes to uncomment this.

leeskelvin (approved):

PR #77, opened 2024-10-18
DM-46942: Raise UpstreamFailureNoWorkFound in ReprocessVisitImageTask when detector is missing a necessary calibration

laurenam:


python/lsst/drp/tasks/reprocess_visit_image.py
```diff
@@ -326,9 +326,22 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
+            msg += "\n  > No entry for the detector was found in the visit summary table"
+        else:
+            if detector_summary.psf is None:
+                msg += "\n  > the PSF model for the detector is None"
+            if detector_summary.wcs is None:
+                msg += "\n  > the WCS model for the detector is None"
+            if detector_summary.apCorrMap is None:
+                msg += "\n  > the aperture correction model map for the detector is None"
+            if detector_summary.photoCalib is None:
+                msg += "\n  > the photometric calibration model for the detector is None"
```
TallJimbo, line 340: It's not a big deal here, but when composing multi-line strings programmatically, it's usually better to append to a `list` of lines and use `"\n".join(lines)`" to stitch them together.

The main advantage is memory allocation efficiency, which only comes into play when there is a very large number of lines, but I think that makes it good practice for consistency everywhere.

python/lsst/drp/tasks/reprocess_visit_image.py
```diff
@@ -326,9 +326,22 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
+        if len(msg) > 0:
```
TallJimbo, line 341: `if msg:` is generally better (or `if lines:`, per the thread above).

tests/test_reprocess_visit_image.py
```diff
@@ -296,12 +305,26 @@ def setUp(self):
...
         self.butler.put(lsst.afw.math.BackgroundList(background), "initial_pvi_background", self.visit_id)
         self.butler.put(lsst.afw.math.BackgroundList(background), "initial_pvi_background", self.visit1_id)
+        self.butler.put(lsst.afw.math.BackgroundList(background), "initial_pvi_background", self.visit2_id)
         self.butler.put(lsst.afw.math.BackgroundList(background), "skyCorr", self.visit_id)
         self.butler.put(lsst.afw.math.BackgroundList(background), "skyCorr", self.visit1_id)
+        self.butler.put(lsst.afw.math.BackgroundList(background), "skyCorr", self.visit2_id)
         self.butler.put(lsst.afw.table.SourceCatalog().asAstropy(), "finalized_src_table", self.visit_only_id)
-        self.butler.put(make_visit_summary(detector=detector), "finalVisitSummary", self.visit_only_id)
+        # Make a simple single gaussian psf so that psf is not None in
+        # finalVisitSummary table whcih would result in
```
TallJimbo, line 317:
```suggestion
        # finalVisitSummary table which would result in
```

TallJimbo (approved):

python/lsst/drp/tasks/reprocess_visit_image.py
laurenam: Always love a good best-practice coding tip from you 😄

PR #80, opened 2024-11-14
DM-47517: Transfer the new warping tasks from pipe_tasks

arunkannawadi:


python/lsst/drp/tasks/make_direct_warp.py
```diff
@@ -0,0 +1,954 @@
+# This file is part of drp_tasks.
```
enourbakhsh, line 1: Thanks for not forgetting to change this line :)

enourbakhsh (approved):
Looks good!

PR #82, opened 2024-12-10
DM-48046: Elevate assemble subregion log to INFO level.

davedavemckay:


python/lsst/drp/tasks/assemble_coadd.py
```diff
@@ -926,7 +926,7 @@ def assembleSubregion(
         nImage : `lsst.afw.image.ImageU`, optional
             Keeps track of exposure count for each pixel.
         """
-        self.log.debug("Computing coadd over %s", bbox)
+        self.log.info("Stacking subregion %s", bbox)
```
yalsayyad, line 929: I'm going to leave blank comment just as an example.

yalsayyad (approved):

PR #88, opened 2025-02-06
DM-47169: Increase threshold to protect only brighter stars

yalsayyad:


laurenam (approved):
It wouldn't hurt to provide a few more details about the decisions made here (either in the commit message or comments in the code).  E.g. "Controls protected zone..." doesn't tell me the "sign" on that, and "Should be >=..." doesn't tell me what the consequences are for leaving this at the higher default.

erykoff:
Why did the temp local background config go away?

yalsayyad:
OK, so I didn't see anything scary, but decided I want to merge a change even more conservative than what I presented in team meeting this week. Just changing the thresholdValue, but leaving the nSigmaToGrow as is.  I'm going to have to do all the runs again with the updated calibs. defect masks that are going in this week anyway.

PR #89, opened 2025-02-06
DM-48790: Clear the mask planes before making PSF-matched warps

arunkannawadi:


leeskelvin (approved):
Thanks Arun, these changes make sense to me, with some nice API cleanup to boot.

PR #90, opened 2025-02-10
DM-46961: Populate the coadd cells along the patch boundary

arunkannawadi:
The scope of this PR is a bit broader than what it started out to be, but I didn't bother to create separate tickets. The commits are however atomic and should be clear in what the changes are.

### DIFF
--- a/python/lsst/drp/tasks/assemble_cell_coadd.py
+++ b/python/lsst/drp/tasks/assemble_cell_coadd.py
@@ -399,9 +399,9 @@ class AssembleCellCoaddTask(PipelineTask):
             reject = afwImage.Mask.getPlaneBitMask(["CLIPPED", "REJECTED"])
             for cellInfo in skyInfo.patchInfo:
                 bbox = cellInfo.outer_bbox
-                mi = warp[bbox].getMaskedImage()
+                mi = warp[bbox].maskedImage
 
-                if (mi.getMask().array & edge).any():
+                if (mi.mask.array & edge).any():
                     self.log.debug(
                         "Skipping %s in cell %s because it has an EDGE bit set",
                         warpRef.dataId,
@@ -409,7 +409,7 @@ class AssembleCellCoaddTask(PipelineTask):
                     )
                     continue
 
-                if (mi.getMask().array & reject).any():
+                if (mi.mask.array & reject).any():
                     self.log.debug(
                         "Skipping %s in cell %s because it has a CLIPPED or REJECTED bit set",
                         warpRef.dataId,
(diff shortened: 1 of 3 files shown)


python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -114,6 +114,15 @@ class AssembleCellCoaddConfig(PipelineTaskConfig, pipelineConnections=AssembleCe
...
         default=True,
     )
+    mask_propagation_thresholds = DictField[str, float](
+        doc=(
+            "Threshold (in fractional weight) of rejection at which we "
+            "propagate a mask plane to the coadd; that is, we set the mask "
+            "bit on the coadd if the fraction the rejected frames "
+            "would have contributed exceeds this value."
+        ),
+        default={"SAT": 0.1, "INEXACT_PSF": 0.0},
```
TallJimbo, line 128: I'm confused as to why `INEXACT_PSF` is in this mapping at all - do we ever reject pixels because of `INEXACT_PSF`?  Is that effectively what's happening when we make a traditional cell-based coadd with no discontinuities?

I'd also have proposed a default for `SAT` over 0.5. Is this a value used in the current coaddition code?

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -399,7 +399,7 @@ def run(self, inputWarps, skyInfo, **kwargs):
                 self.scale_zero_point.run(exposure=warp, dataRef=warpRef)
 
             # Coadd the warp onto the cells it completely overlaps.
-            edge = afwImage.Mask.getPlaneBitMask("EDGE")
+            edge = afwImage.Mask.getPlaneBitMask("NO_DATA")
```
TallJimbo, line 402: Might be more readable to change the variable name, too.

python/lsst/drp/tasks/assemble_cell_coadd.py
arunkannawadi: Ah, this is leftover from an earlier version of this ticket when I wanted to `INEXACT_PSF` mask bits sets. We've since moved away from that approach and will remove it. FWIW, unless mentioned here, all mask planes, whether defined or not, have a default threshold value of 0.0.

The value for `SAT` is based on what it is currently used in `assemble_coadd` but I don't have strong opinions on what the value should be.

TallJimbo: Ok, let's leave the SAT value the same as `assemble_coadd` for now and revisit it later, then.

PR #93, opened 2025-02-18
DM-48649: Use per-detector weights for cell-based coadds

arunkannawadi:


python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -76,6 +82,15 @@ class AssembleCellCoaddConnections(
...
     )
 
+    visitSummaryList = Input(
+        doc="Input visit-summary catalogs with updated calibration objects. Mainly used for coadd weights.",
+        name="finalVisitSummary",
+        storageClass="ExposureCatalog",
+        dimensions=("instrument", "visit"),
+        deferLoad=True,
+        multiple=True,
+    )
```
TallJimbo, line 92: One potential downside here is that it opens up `AssembleCellCoadd` to the Piff memory leak.  I'm not sure if that's a problem at all, but since it'd be all PSFs for all detectors in any visit being coadded, it might be a larger number of those PSFs in memory than we typically have in a single quantum.

An alternative might be the parquet `ccdVisitTable`.  Thus far we haven't used that as an input to anything other than analysis tasks, but I don't think there's any reason we couldn't.

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -391,18 +429,47 @@ def run(self, inputWarps, skyInfo, **kwargs):
...
+            else:
+                zero_point_scale_factor = 1.0
 
             # Coadd the warp onto the cells it completely overlaps.
             edge = afwImage.Mask.getPlaneBitMask(["NO_DATA", "SENSOR_EDGE"])
             reject = afwImage.Mask.getPlaneBitMask(["CLIPPED", "REJECTED"])
             removeMaskPlanes(warp.mask, self.config.remove_mask_planes, self.log)
 
+            # Compute the weight for each CCD in the warp from the visitSummary
+            # or from the warp itself, it not provided. Computing the weight
```
TallJimbo, line 450:
```suggestion
            # or from the warp itself, if not provided. Computing the weight
```

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -391,18 +429,47 @@ def run(self, inputWarps, skyInfo, **kwargs):
...
                 del artifactMask
 
-            # Pre-process the warp before coadding.
             # Each Warp that goes into a coadd will typically have an
             # independent photometric zero-point. Therefore, we must scale each
             # Warp to set it to a common photometric zeropoint.
             if self.config.do_scale_zero_point:
-                self.scale_zero_point.run(exposure=warp, dataRef=warpRef)
+                imageScaler = self.scale_zero_point.run(exposure=warp, dataRef=warpRef).imageScaler
+                zero_point_scale_factor = imageScaler.scale
```
TallJimbo, line 437: Are we certain that this is going to be a scalar `ImageScalar` rather than a `SpatialImageScaler` because we're doing it cell-by-cell and evaluating the `PhotoCalib` only at the centers of cells?  If it's a `SpatialImageScaler` it won't have your new `scale` property, because there won't be a single number, and I don't think we can assume the `PhotoCalib` will be spatially constant.

TallJimbo (approved):

python/lsst/drp/tasks/assemble_cell_coadd.py
arunkannawadi: I ran a test run over several hundred patches without any trouble, so I doubt that this is an issue. But `ccdVisitTable` was my first thought but changed it to `finalVisitSummary` after we discussed about it. I'd be happy to switch it to `ccdVisitTable` and may be even read just the column corresponding to `meanVar`.

python/lsst/drp/tasks/assemble_cell_coadd.py
arunkannawadi: Hmm, good point. This works on the current defaults, but let me look into what the best way to obtain a scalar value would be in all cases.

python/lsst/drp/tasks/assemble_cell_coadd.py
arunkannawadi: Relying on a parquet table for weights/mean variance has the advantage that I can mock this up easliy in unit tests, and it's possible for simulations running this code outside of normal pipeline/middleware workflow to provide in a replacement.

python/lsst/drp/tasks/assemble_cell_coadd.py
arunkannawadi: Given the [100+ post thread](https://rubin-obs.slack.com/archives/C07TD4VUJ3X/p1740075497308109) we just had, I am even more certain that we shouldn't worry too much to generalize this to work with `SpatialImageScaler`. That, along with `SpatialScaleZeroPointTask` is used nowhere in our code base, not even in unit tests. I'll wait for @fred3m 's changes to come in before I decide how to proceed here.

TallJimbo: I completely agree.  I hadn't realized the spatial version was going unused before.

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -87,9 +102,21 @@ class AssembleCellCoaddConnections(
         doc="Output multiple cell coadd",
         name="{inputWarpName}Coadd{outputCoaddSuffix}",
         storageClass="MultipleCellCoadd",
-        dimensions=("tract", "patch", "band", "skymap"),
+        dimensions=("tract", "patch", "band", "skymap", "instrument"),
```
arunkannawadi, line 105: I don't think we can add `instrument` just yet unless we clean up the existing datasets, but this is something we want to add soon(-ish).

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -221,8 +257,9 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
             skyMap, tractId=outputDataId["tract"], patchId=outputDataId["patch"]
         )
 
+        units = CoaddUnits.legacy if self.config.do_scale_zero_point else CoaddUnits.nJy
```
arunkannawadi, line 260: The value of `do_scale_zero_point` is essentially used to identify whether the warps are in units of `nJy` or not. Given that we haven't used this for any serious purpose yet, we _could_ drop support to make cell-coadds from old warps but I have retained that capability now although in a less robust manner.

arunkannawadi (commented):
Re-requesting review on this ticket since setting the pixel units to `nJy` has changed a few things here.

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -397,18 +436,56 @@ def run(self, inputWarps, skyInfo, **kwargs):
...
+
+            if visitSummaryRef:
+                assert visitSummaryRef.dataId["visit"] == warpRef.dataId["visit"]
+                visitSummary = visitSummaryRef.get()
+                for detector in full_ccd_table["ccd"]:
+                    visitSummaryRow = visitSummary.find(detector)
+                    mean_variance = visitSummaryRow["meanVar"]
+                    mean_variance *= zero_point_scale_factor**2
+                    if warp.metadata.get("BUNIT", None) == "nJy":
+                        mean_variance *= visitSummaryRow.photoCalib.getCalibrationMean() ** 2
```
arunkannawadi, line 477: The need for the `PhotoCalib` object here means we can't use `ccdVisitTable` in place of `finalVisitSummary`; unless we put this info there

TallJimbo: I think loading the visit summary here is probably better than either calibrating the summary stats quantities or putting the calibration mean into the CCD-visit table.

TallJimbo (approved):

python/lsst/drp/tasks/assemble_cell_coadd.py
arunkannawadi: @timj had mentioned about not having that info in the metadata and having to rely only the collection name. My understanding was not having this for `deepCoadd` was a mistake but something that we can fix for the cell-coadds.

https://rubin-obs.slack.com/archives/C08CANY4B6H/p1740701138145739?thread_ts=1740699057.980539&cid=C08CANY4B6H

arunkannawadi: If we do want to add `instrument` now, that still leaves us with a question with one of the three possible answers:

1. Change the dataset type name
2. Delete all the `deepCoaddCell` datasets that Miranda and I have made, deregister and re-register the dataset type.
3. Do some database surgery to change the dimensions without affecting anything else (this is beyond me, and I don't know how easy it is)

TallJimbo: Tim may well win this debate in the end, but let's not do it on this PR since that does create dataset type name clashes.

Instead, let's try to get this right before we put the cell-based coadds in a DRP-v2 pipeline, since we'd want to change the dataset type name there anyway.

arunkannawadi: Sounds good to me.

PR #94, opened 2025-02-18
DM-48783: Add illumination correction support to ReprocessVisitImageTask

erykoff:


TallJimbo (approved):
I think `UpdateVisitSummaryTask` does some summary-stats stuff with the background as well.  Can you take a look to see if it needs changes, too?

erykoff:
Looking at the `UpdateVisitSummaryTask` to goes back to `pipe_tasks` `ComputeExposureSummaryStatsTask`.  The only thing there that might be relevant is the sky stats on the image.  And now I'm not sure about that.  Thoughts?

TallJimbo (approved):

PR #104, opened 2025-03-25
DM-49675: Set ref cat epoch at config level.

PFLeget:
See thread [here](https://rubin-obs.slack.com/archives/C07TD4VUJ3X/p1742485414070239)

### DIFF
--- a/python/lsst/drp/tasks/gbdesAstrometricFit.py
+++ b/python/lsst/drp/tasks/gbdesAstrometricFit.py
@@ -450,6 +450,12 @@ class GbdesAstrometricFitConfig(
         "returned are not used.",
         default="phot_g_mean",
     )
+    setRefEpoch = pexConfig.Field(
+        dtype=float,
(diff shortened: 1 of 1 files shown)


python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -450,6 +450,12 @@ class GbdesAstrometricFitConfig(
         "returned are not used.",
         default="phot_g_mean",
     )
+    setRefEpoch = pexConfig.Field(
+        dtype=float,
+        doc="Set reference epoch (set to median if None). (units of input in mjd)",
```
ctslater, line 455: I'm trying to think of how to make description clear, how about: "Set the reference epoch to a fixed value in MJD (if None, median observation date is used)" ?

ctslater (approved):
Looks good, minor docstring suggestion.

PR #106, opened 2025-03-31
DM-49343: Persist Dcr metric convergence from DcrAssembleCoadd

abudlong:


bsmartradio (approved):
Everything looks good, assume Jenkins passed with no issues.

python/lsst/drp/tasks/dcr_assemble_coadd.py
```diff
@@ -1047,6 +1065,48 @@ def calculateConvergence(self, dcrModels, subExposures, bbox, warpRefList, weigh
         self.log.info("Individual metrics:\n%s", metricList)
         return 1.0 if weight == 0.0 else metric / weight
 
+    def calculateTotalConvergence(self, dcrModels, bbox, warpRefList, weightList, statsCtrl):
```
isullivan, line 1068: Please find a way to re-use `calculateConvergence` with slight modification rather than duplicating the code. For example, you could call `exposure = warpExpRef.get()` if `subExposures` is `None`, and use `exposure = subExposures[visit][bbox]` if it is not `None`.

isullivan (approved):

PR #109, opened 2025-04-21
DM-50358: Pass padding parameter when constructing the grid

arunkannawadi:


python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -314,8 +314,13 @@ def _construct_grid(skyInfo):
         """
         # grid has no notion about border or inner/outer boundaries.
         # So we have to clip the outermost border when constructing the grid.
-        grid_bbox = skyInfo.patchInfo.outer_bbox.erodedBy(skyInfo.patchInfo.getCellBorder())
-        grid = UniformGrid.from_bbox_cell_size(grid_bbox, skyInfo.patchInfo.getCellInnerDimensions())
+        padding = skyInfo.patchInfo.getCellBorder()
```
TallJimbo, line 317: Is the comment above the changed line still correct?

TallJimbo (approved):

python/lsst/drp/tasks/assemble_cell_coadd.py
arunkannawadi: It is not. Good catch,l et me change that.

PR #112, opened 2025-04-25
DM-47521: Refer to makeDirectWarp from drp_tasks

arunkannawadi:
When the docs moved from `pipe_tasks` to `drp_tasks`, these should have been renamed but weren't. Now that they are removed in `pipe_tasks`, not redirecting this breaks the documentation build.


arunkannawadi:
Single package docs builds successfully with this branch (witha  ton of warnings unrelated to the changes here). Merging now.

PR #116, opened 2025-05-09
DM-50735: Add new task for fitting visit-level backgrounds.

TallJimbo:


python/lsst/drp/tasks/fit_visit_background.py
```diff
@@ -0,0 +1,674 @@
...
+        detector_slices : `~collections.abc.Mapping` [`int`, `slice`]
+            Mapping from detector ID to the slice that corresponds to that
+            detector in ``x`` and ``y``.
+        detectors : `~collections.abc.Mapping` [`int`,\
+                `lsst.afw.cameraGeom.Detector`]
+            Descriptions of all detectors that the matrix will evaluate on.
+
+        Returns
+        -------
+        matrix : `numpy.ndarray
```
erykoff, line 317: Missing backtick

erykoff (approved):
This looks good, thanks for the walk-through.  I guess the real proof is in the application.  A couple of small typos.  My only larger comment is that I found `detector_slices` to be confusing nomenclature, but I don't have a better suggestion.  In these cases I explicitly use indices rather than slices but maybe that's just because of how I think about this sort of thing.

PR #129, opened 2025-06-16
DM-51332: Fix where BUNIT is set in reprocessVisitImage

TallJimbo:


python/lsst/drp/tasks/reprocess_visit_image.py
```diff
@@ -148,7 +148,7 @@ class ReprocessVisitImageConfig(
     )
     remove_initial_photo_calib = pexConfig.Field(
         dtype=bool,
-        default=True,
+        default=False,
```
erykoff, line 151: Is this False because default DRP pipelines don't apply the photocalib in CalibrateImageTask?

erykoff (approved):

PR #133, opened 2025-07-08
DM-47858: Add SingleFrameDetectAndMeasureTask

isullivan:
This is a stripped-down version of reprocessVisitImage that can be used to detect sources on preliminary_visit_images

### DIFF
--- a/python/lsst/drp/tasks/single_frame_detect_and_measure.py
+++ b/python/lsst/drp/tasks/single_frame_detect_and_measure.py
@@ -1,0 +1,298 @@
+# This file is part of drp_tasks.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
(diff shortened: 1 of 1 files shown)


python/lsst/drp/tasks/single_frame_detect_and_measure.py
```diff
@@ -0,0 +1,298 @@
...
+    )
+
+    def setDefaults(self):
+        super().setDefaults()
+
+        # Re-estimate the background
+        self.detection.reEstimateBackground = True
+        self.detection.doTempLocalBackground = False
+
+        # NOTE: these apertures were selected for HSC, and may not be
```
erinleighh, line 129: I know this is also a comment on `ReprocessVisitImageTask`, but would it be worth making this a todo with an investigation ticket?

erinleighh (approved):
Everything looked good, thank you!

PR #134, opened 2025-07-21
DM-51882: Ensure coordinate error fields are present in single frame schema

isullivan:


python/lsst/drp/tasks/single_frame_detect_and_measure.py
```diff
@@ -185,7 +185,7 @@ def __init__(self, schema=None, **kwargs):
             type="U",
             doc="Detector this source appeared on.",
         )
-
+        afwTable.CoordKey.addErrorFields(self.schema)
```
enourbakhsh, line 188: I think you meant `afwTable.CoordKey.addErrorFields(schema)`?

enourbakhsh (changes requested):
Just a minor comment, otherwise looks good.

python/lsst/drp/tasks/single_frame_detect_and_measure.py
isullivan: Good catch! Thank you.

PR #138, opened 2025-08-07
DM-52110: Add metadetect as an optional dependency

arunkannawadi:


arunkannawadi:
The idea is to keep the metadetect `PipelineTask` here in `drp_tasks`. Making `metadetect` package use `Task` framework (so we can use them as `subTask`s here) was already a struggle and making it a `PipelineTask` there would be even more difficult.

PR #142, opened 2025-09-03
DM-51641: Fix AnnotatedPartialOutputsError in MakePsfMatchedWarpTask and add a test

fred3m:


tests/test_make_psf_matched_warp.py
```diff
@@ -69,6 +89,18 @@ def test_makeWarp(self):
         # Ensure the warp has valid pixels
         self.assertGreater(np.isfinite(psf_matched_warp.image.array.ravel()).sum(), 0)
 
+        # Test that a failed PSF match generates
```
erykoff, line 92: I think this should all be it's own `def test_annotated_partial_outputs()` or some such.  I think that `warp` can be replaced with any object in the mock?

erykoff (approved):
Minor comment, may be a bit cleaner.

PR #144, opened 2025-09-09
DM-52305: Address out of range error with cell-coadds

arunkannawadi:


erykoff (approved):
Can we also write a test for this?  Presumably at least you've run the failing dataId, but it's better to test directly (we've had these AlgorithmErrors crash in use before...)

PR #146, opened 2025-09-16
DM-52525: Fix crash when there are no best wcs matches.

erykoff:


python/lsst/drp/tasks/update_visit_summary.py
```diff
@@ -312,10 +312,13 @@ def best_for_detector(
         results = np.array(results)
         distances = np.array(distances)
         if inPixel.sum() == 1:
+            # One match, we return it.
             best_result = results[inPixel][0]
         elif inPixel.sum() == 0:
-            best_result = results[distances.argmin()]
+            # We did not find any matches.
+            best_result = [-1, None]
```
cmsaunders, line 319: We don't want to do this. These three options are here because the healpix Regions do not perfectly describe the actual regions, so a given point can fall between two "Regions", or be "contained" by two (not 100% sure both of these cases occur, but one of them does), even though it should actually be in exactly one of them. 

I think this error is occurring for a separate reason, which is that no `wcs_overrides_skypix` were available for this visit. We need a separate treatment for cases where `self.catalogs_by_pixel` is empty.

cmsaunders (commented):
See my comment below. This needs to be solved a different way.

PR #215, opened 2026-02-24
DM-54193: Allow for case when there are more backgrounds than images

yalsayyad:
This is possible if creating warps from
visit_image and using skyCorr which is generated
from preliminary_visit_image.

### DIFF
--- a/python/lsst/drp/tasks/make_direct_warp.py
+++ b/python/lsst/drp/tasks/make_direct_warp.py
@@ -432,11 +432,23 @@ class MakeDirectWarpTask(PipelineTask):
         # they're present for the same detectors we got image handles for, in
         # case of upstream errors.
         for ref in getattr(inputRefs, "background_revert_list", []):
-            inputs[ref.dataId["detector"]].background_revert = butlerQC.get(ref)
+            detector_id = ref.dataId["detector"]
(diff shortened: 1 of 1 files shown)


python/lsst/drp/tasks/make_direct_warp.py
```diff
@@ -432,11 +432,23 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
         # they're present for the same detectors we got image handles for, in
         # case of upstream errors.
         for ref in getattr(inputRefs, "background_revert_list", []):
-            inputs[ref.dataId["detector"]].background_revert = butlerQC.get(ref)
+            detector_id = ref.dataId["detector"]
+            if detector_id in inputs:
+                inputs[detector_id].background_revert = butlerQC.get(ref)
+            else:
+                self.log.info(f"Got {ref.datasetType.name} for detector {detector_id} but no image")
```
arunkannawadi:
```suggestion
                self.log.info("Got %s for detector %d but no image", ref.datasetType.name, detector_id)
```

arunkannawadi: and similarly in the log messages below.

arunkannawadi (approved):

PR #219, opened 2026-04-09
DM-54418: Update stellar motions covariance output

cmsaunders:


fred3m (approved):
Everything that you've done looks good but I wonder if it would be useful to include a helper function to extract the appropriate columns from the catalog to rebuild the covariance matrix. I'll leave it up to your discretion whether or not that would be worth implementing. If so, then you probably want to make a constant that contains the `positionNames` in case they ever change. In fact, you might want to do that anyway instead of having them hard coded inside of the method.

PR #221, opened 2026-04-21
DM-54708: Stop mutating input argument to MakeDirectWarpTask.run.

TallJimbo:


python/lsst/drp/tasks/make_direct_warp.py
```diff
@@ -879,21 +880,22 @@ def _apply_all_calibrations(
 
     # This method is copied from makeWarp.py
```
taranu: This used to exist in pipe_tasks, right? And no longer does?

python/lsst/drp/tasks/make_direct_warp.py
```diff
@@ -879,21 +880,22 @@ def _apply_all_calibrations(
...
     @classmethod
-    def _prepareEmptyExposure(cls, sky_info):
+    def _prepareEmptyExposure(cls, bbox, wcs):
```
taranu, line 871: Now that we've upgraded documenteer, can we finally just add type hints here instead of only in docstrings? Or does there need to be a mass change in every package?

taranu (commented):
The code changes look fine. My main comment is there seem to be a fair number of references to the old makeWarp which appears to be gone for good now, so at some point those should be removed/updated to be more relevant.

python/lsst/drp/tasks/make_direct_warp.py
TallJimbo: Correct.  I've removed this comment and the bit in the class docstring (some of which had actually become incorrect since it was written).

python/lsst/drp/tasks/make_direct_warp.py
TallJimbo: I think if we add type hints here it will just work in the `pipelines.lsst.io` build, but we do need per-package modifications to the `doc` directory to get `package-docs build` working again for development (though I think that may already be broken).

I hadn't before considered adding type annotations just for docs (i.e. without any type-checking actually running).  I'm not sure how I feel about it; it's a step in the right direction (or at least I think so), but I have some concern about people (and especially their editors) thinking they're more rigorous than they actually are.

PR #222, opened 2026-05-06
DM-54870: Fix the buffer bug

arunkannawadi:


python/lsst/drp/tasks/metadetection_shear.py
```diff
@@ -634,6 +689,12 @@ def runQuantum(
     ) -> None:
         # Docstring inherited.
 
+        # Get the skyMap for this quantum
+        sky_map = qc.get(inputRefs.sky_map)
+
+        if self.config.do_validate_skymap_config:
+            self.validate_skymap_config(sky_map.config)
```
arunkannawadi: The validation that happens here should ideally happen _once_ at the config time, and not once per quanta. But I don't know if we can get two configs and run these checks. Having the `do_validate_skymap_config` config parameter gives us an option of turning off those checks, but they are done by default.

python/lsst/drp/tasks/metadetection_shear.py
```diff
@@ -626,6 +640,47 @@ def make_metadetect_schema(cls, config: MetadetectionShearConfig) -> pa.Schema:
 
         return pa_schema
 
+    def validate_skymap_config(self, skymap_config: BaseSkyMap.ConfigClass) -> None:
+        if not skymap_config.tractBuilder.name == "cells":
+            raise InvalidQuantumError("MetadetectionShearTask requires a cell-based skymap.")
+
+        cell_config = skymap_config.tractBuilder.active
+        if (self.config.border is None and cell_config.cellBorder == 0) or (
```
TallJimbo: Is allowing `self.config.border == 0 and cell_config.cellBorder == 0` intentional here?

This also makes me wonder if allowing `self.config.border=None` is useful here; doesn't seem to be materially different from zero (except in this check).

python/lsst/drp/tasks/metadetection_shear.py
```diff
@@ -626,6 +640,47 @@ def make_metadetect_schema(cls, config: MetadetectionShearConfig) -> pa.Schema:
...
+                    f"Maximum border value is {max_border_value}."
+                )
+
+            if (
+                skymap_config.tractOverlap * 3600 / skymap_config.pixelScale
+                < self.count_cells_along_edges(skymap_config) * cell_inner_dimensions + cell_config.cellBorder
+            ):
+                raise InvalidQuantumError(
+                    "The pixel scale in the skymap config does not match the pixel scale in the task config."
+                )
```
TallJimbo, line 672: Where does a pixel scale in the task config appear here?  I'm not sure what this check is trying to do, unless it's verifying the skymap's own invariants.

python/lsst/drp/tasks/metadetection_shear.py
TallJimbo: There's no good way to do that only once; the system is designed instead for the theoretical case where you could actually get data IDs with different skymaps in the same QG (not because we need that, but because it's analogous to the case where there are multiple instruments in QG, and that seems useful).

In any case, the check seems like it should be quite fast, so I don't think there's anything to worry about here.

python/lsst/drp/tasks/metadetection_shear.py
```diff
@@ -704,12 +769,31 @@ def run(
         self.rng = np.random.RandomState(seed)
         idstart = 0
 
+        # We bake in the reasonable assumption that the skymap is configured
+        # to have one cell as the patch border. This should be ideally read
+        # from the skymap config with an attribute of the same name.
```
TallJimbo: This comment was obsoleted by a later commit.

TallJimbo (approved):

python/lsst/drp/tasks/metadetection_shear.py
arunkannawadi: The logic is correct, but the error message is not. It should read "The tract overlap is insufficient and will result in missed regions between adjacent tracts" or something like that. It is implicitly assumed that `tractOverlap` is in degrees, while everything else is in pixel units.

python/lsst/drp/tasks/metadetection_shear.py
arunkannawadi: Right. I thought the second commit removed these lines, but it apparently didn't. I'll remove this comment.

python/lsst/drp/tasks/metadetection_shear.py
arunkannawadi: Allowing `self.config.border` to be None clarifies that the user intended to rely on the `cellBorder` defined in the `skymap` to create the buffer region, and not the `border` value put in by hand. Allowing `self.config.border == 1 and cell_config.cellBorder == 0` would be just as bad as `self.config.border == 0 and cell_config.cellBorder == 0`. The minimum allowed buffer region would depend on many factors (PSF size, artificial metacal shear applied, cell size etc.) and we are not mandating that specified `border` is larger than that minimum value.

arunkannawadi: OK, I see your point now. The case I absolutely want to avoid is the one where two (different) positive values are set for `cellBorder` and `self.config.border`. As long as one of them is a positive integer and the other is not, that is a valid configuration to accept.

python/lsst/drp/tasks/metadetection_shear.py
arunkannawadi: I have removed the `do_validate_skymap_config` config field. One of my intentions was also to have the option to remove the connections for skymap if this was set to `False`, but we need it for calculation anyways.

python/lsst/drp/tasks/metadetection_shear.py
arunkannawadi: `border` cannot be set to `None`.

PR #223, opened 2026-05-07
DM-53026: Add option to use isolated stars as input

cmsaunders:


tests/test_gbdesAstrometricFit.py
```diff
@@ -256,8 +387,8 @@ def _make_refCat(cls, starIds, starRas, starDecs, inReferenceFraction, bounds):
             record.setRa(lsst.geom.Angle(starRas[i], lsst.geom.degrees))
             record.setDec(lsst.geom.Angle(starDecs[i], lsst.geom.degrees))
             record.set(fluxKey, 1)
-            record.set(raErrKey, 0.00001)
-            record.set(decErrKey, 0.00001)
+            record.set(raErrKey, 1e-8)
```
kherner, line 390: Probably doesn't make any difference, but I was curious about the rationale for this change of the err values.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -444,6 +462,66 @@ class GbdesAstrometricFitConnections(
...
+        # the inputs of each quantum.
+        for quantum_data_id, visit_data_ids_for_quantum in visits_by_quantum.items():
+            # We'll use tracts_for_quantum to track which inputs are already
+            # present.
+            tracts_for_quantum = tracts_by_quantum[quantum_data_id]
+            for visit_data_id in visit_data_ids_for_quantum:
+                for tract_data_id in tracts_by_visit[visit_data_id]:
+                    if tract_data_id not in tracts_for_quantum:
+                        adjuster.add_input(quantum_data_id, "isolatedStarCatalogs", tract_data_id)
+                        adjuster.add_input(quantum_data_id, "isolatedStarSources", tract_data_id)
```
kherner, line 522: There's never any need to go the other direction (delete something already there), right? I don't think so, since a slightly larger than necessary catalog probably doesn't hurt anything.

python/lsst/drp/tasks/gbdesAstrometricFit.py
```diff
@@ -2641,7 +2996,7 @@ def run(
         )
 
         # Add the science and reference sources
-        self._add_objects(wcsf, sourceDict, extensionInfo)
+        self._add_objects_from_dict(wcsf, sourceDict, extensionInfo)
```
kherner, line 2999: Do you need a switch here for whether you're running doIsolatedStars or not? Something like

```
if not config.doIsolatedStars:
    self._add_objects(wcsf, sourceDict, extensionInfo)
else:
    self._add_objects_from_dict(wcsf, sourceDict, extensionInfo)
```

Maybe it's taken care of by other stuff upstream; it was a bit difficult for me to follow.

kherner (approved):
I think everything is OK, but I had a couple of minor conceptual questions and one thing to double check whether you need an if-else or not.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: No, this is in the inherited `GbdesGlobalAstrometricFit` class, which only ever used isolated stars. Using isolated stars as input is the "preferred" method, but until `_adjust_all_quanta` was created, it wasn't possible to use them in the base class. The `_add_objects` method was overload in the inherited class, but now it has been moved into the base class, so I had to give it a new name.

tests/test_gbdesAstrometricFit.py
cmsaunders: This actually makes a huge difference! I should have commented on this on Jira. In the course of adding a test for the isolated stars input, I realized that the mock reference stars had errors that were way too high, which was preventing the model from fitting isolated stars well at the edges of the focal plane. It took me literally days to figure out the problem. However, with this change, _all_ the tests actually work better and run faster, so I don't feel too bad about the time spent on it.

python/lsst/drp/tasks/gbdesAstrometricFit.py
cmsaunders: I think there might be some cases where the butler gives you tracts that overlap with the healpix of the quantum, but that don't actually overlap with any of the input visits, but they will be filtered out later. I think that process is fast enough not to worry about it.

python/lsst/drp/tasks/gbdesAstrometricFit.py
kherner: Thanks; I did not appreciate it was in the other class. I should have expanded the auto-hidden stuff above it and I would have understood better.

PR #228, opened 2026-05-21
DM-54947: Fix assembleCellCoadd handling of fully-masked cells

TallJimbo:


python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -782,6 +783,23 @@ def run(
                     )
                     continue
 
+                # Compute the unmasked fraction for this detector in the inner
+                # cell. Used to gate on max_maskfrac and to scale the PSF weight.
+                inner_detector_pixels = detector_map[inner_bbox].array == ccd_row["ccd"]
+                inner_bad_pixels = (warp[inner_bbox].mask.array & rejected) != 0
+                unmasked_fraction = (
+                    (inner_detector_pixels & ~inner_bad_pixels).sum() / inner_detector_pixels.sum()
+                )
```
TallJimbo: This would read better as
```
inner_unmasked_pixels = (warp[inner_bbox].mask.array & rejected) == 0
unmasked_fraction = (
    (inner_detector_pixels & inner_masked_pixels).sum() / inner_detector_pixels.sum()
)
```
(i.e. just avoiding a double negative)

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -782,6 +783,23 @@ def run(
...
+                if unmasked_fraction < (1.0 - self.config.max_maskfrac):
```
TallJimbo:
```suggestion
                if unmasked_fraction <= max(1.0 - self.config.max_maskfrac, 0.0):
```
I think the weighting below handles the the case where `max_maskfrac==1.0` and `unmasked_fraction=0.0`, but I think it's even better if we just `continue` here in that case (less work, less chance of new bugs downstream).

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -845,7 +864,10 @@ def run(
                         psf_eval_point,
                     )
 
-                overlaps_center = detector_map[geom.Point2I(bbox.getCenter())] == ccd_row["ccd"]
+                psf_weight = weight * unmasked_fraction
+                overlaps_center = (
+                    detector_map[geom.Point2I(bbox.getCenter())] == ccd_row["ccd"]
+                ) and (psf_weight > 0.0)
```
TallJimbo: Applying `unmasked_fraction` to the PSF weight wasn't actually something I considered, but I think it's a good idea.  I'm interested in @arunkannawadi's thoughts.  It's clearly better than not modifying the weights when `unmasked_fraction << 1`, and it's not obviously worse than leaving them alone in `unmasked_fraction` is larger.

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -854,6 +876,7 @@ def run(
                 observation_identifiers_gc[cellInfo.index][observation_identifier] = CoaddInputs(
                     overlaps_center=overlaps_center,
                     overlap_fraction=overlap_fraction,
+                    unmasked_overlap_fraction=unmasked_fraction,
```
TallJimbo, line 921: :+1: to this addition, but I suspect this alone is not going to be enough to get it saved; I think `CoaddInputs` is a `dataclass`, not a `Struct`, and there's definitely logic that would be needed in another file or two to stuff it into a binary table and read it back out.

jrmullaney: I _think_ this has already been addressed. I had to update `lsst.cell_coadds.CoaddInputs` to allow it to accept `unmasked_overlap_fraction`. But then it failed a whole bunch of tests, at which point I realised I needed to also edit `lsst.cell_coadds._fits`. Those changes are on the same ticket branch in `cell_coadds`.

But I realise that there may be other places in the stack that read in CoaddInputs that may need updating. I guess Jenkins should pick those up?

TallJimbo: Maybe you forgot to `git add` those, or they ended up on the other branch?  I don't see any changes in `_fits` on this PR.

jrmullaney: It was still "just" a branch. PR on in `cell_coadds` repo now made.

TallJimbo: :man_facepalming: I totally forgot we were not already in that package.

TallJimbo (commented):
I can't technically Approve this because I made the PR. But we can consider it reviewed and approved.

I would like to run some data through before we merge, though.

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -800,10 +830,11 @@ def run(
                 warp_stacker_gc[cellInfo.index].add_masked_image(mi, weight=weight)
 
                 if masked_fraction_image:
-                    mi = afwImage.ImageF(masked_fraction_image[bbox], deep=deep_copy)
+                    mi = afwImage.ImageF(masked_fraction_image[bbox], deep=True)
```
arunkannawadi: We should include the case when in-place modification might happen to the `deep_copy` calculation so we don't always make a deep copy.

python/lsst/drp/tasks/assemble_cell_coadd.py
arunkannawadi: Coadding PSFs by a different weight, weighted by `unmasked_fraction` would be incorrect; as an example, if several visits all overlapping the center but having different unmasked fraction would result in the incorrect PSF at the center of the cell. It might work well enough on average, but this is a big change to how we coadd PSFs and I don't think that's a good idea at this stage.

python/lsst/drp/tasks/assemble_cell_coadd.py
TallJimbo: For correctness, or just efficiency?  If the latter, I think that might be premature optimization (I wonder if all of the `deep_copy` logic is), and I don't want to add any more complexity to the logic unless we're sure we need it.

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -782,6 +783,35 @@ def run(
...
+                    self.log.debug(
+                        "Skipping %s in cell %s: masked fraction %.3f exceeds threshold %.3f",
+                        warp_input.dataId,
+                        cellInfo.index,
+                        1.0 - unmasked_fraction,
+                        self.config.max_maskfrac,
+                    )
+                    continue
+
+                psf_weight = weight * unmasked_fraction
```
TallJimbo: GitHub seems to be confused about where to hang @arunkannawadi 's comment on an older version, and won't let me reply directly, so:

> Coadding PSFs by a different weight, weighted by unmasked_fraction would be incorrect; as an example, if several visits all overlapping the center but having different unmasked fraction would result in the incorrect PSF at the center of the cell. It might work well enough on average, but this is a big change to how we coadd PSFs and I don't think that's a good idea at this stage.

Very good point.  I don't think the center pixel argument is particularly compelling (it's not special) but pixels that don't have `INEXACT_PSF` *are* special, in that we do want the PSF model to be exact for those.  @jrmullaney and I had been thinking that this was a choice between getting the PSF more wrong on fewer pixels vs. getting it a little wrong on more pixels.  But we have a way to label the badness (i.e. `INEXACT_PSF`) in the first case and not on the latter.  So yeah, we should not be weighing the PSF by `unmasked_fraction`.

python/lsst/drp/tasks/assemble_cell_coadd.py
arunkannawadi: For efficiency purposes only. I remember changes to the HSM code slowing down significantly because we made a deep copy for every object. This doesn't scale by the number of objects, so it might be fine even in terms of efficacy.

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -782,6 +783,35 @@ def run(
...
                     continue
 
+                # Compute the unmasked fraction for this detector in the inner
+                # cell. Used to gate on max_maskfrac and to scale PSF weight.
+                inner_detector_pixels = detector_map[inner_bbox].array == ccd_row["ccd"]
+                inner_unmasked_pixels = (warp[inner_bbox].mask.array & rejected) == 0
+                unmasked_fraction = (
+                    inner_detector_pixels & inner_unmasked_pixels
+                ).sum() / inner_detector_pixels.sum()
+                if unmasked_fraction <= max(1.0 - self.config.max_maskfrac, 0.0):
```
taranu, line 805: This max isn't necessary since 0 <= max_maskfrac <= 1 is already guaranteed by the RangeField.

python/lsst/drp/tasks/assemble_cell_coadd.py
```diff
@@ -790,15 +801,34 @@ def run(
                 unmasked_fraction = (
                     inner_detector_pixels & inner_unmasked_pixels
                 ).sum() / inner_detector_pixels.sum()
+                is_fallback = is_fallback_gc[cellInfo.index]
                 if unmasked_fraction <= max(1.0 - self.config.max_maskfrac, 0.0):
                     self.log.debug(
-                        "Skipping %s in cell %s: masked fraction %.3f exceeds threshold %.3f",
+                        "Skipping %s in cell %s: masked fraction %.3f exceeds threshold %.3f%s",
```
arunkannawadi: I'd move this logging in the `if not is_fallback` block below.

python/lsst/drp/tasks/assemble_cell_coadd.py
TallJimbo: Yeah, this is probably more paranoid than defensive, but I'm okay with being paranoid on this issue.

arunkannawadi (approved):

PR #229, opened 2026-05-24
DM-55017: Add accommodations for crowded fields in coadd assembly and detection

laurenam:


python/lsst/drp/tasks/assemble_coadd.py
```diff
@@ -1769,9 +1777,39 @@ def findArtifacts(self, templateCoadd, warpRefList, imageScalerList):
...
-            templateFootprints = self.detectTemplate.detectFootprints(sacrificeToDetection)
+            for iFactorNeg in range(self.config.maxDetectTemplateFactorNegIter):
+                try:
+                    templateFootprints = self.detectTemplate.detectFootprints(
+                        sacrificeToDetection, factor=factor, factorNeg=factorNeg
+                    )
+                    break
+                except TooManyMaskedPixelsError as e:
+                    if iFactorNeg < self.config.maxDetectTemplateFactorNegIter - 1:
+                        factorNeg *= 1.25
```
kherner: Pardon my ignorance, but Is there anything special about the choice of 1.25? If not, would it make sense to make it configurable?

python/lsst/drp/tasks/assemble_coadd.py
```diff
@@ -1769,9 +1777,39 @@ def findArtifacts(self, templateCoadd, warpRefList, imageScalerList):
...
+                            "This is likely a result of background over-subtraction leading to a high "
+                            "DETECTED_NEGATIVE masked fraction. Increasing the negative detection factor "
+                            "to %.2f and retrying.",
+                            e,
+                            iFactorNeg + 1,
+                            self.config.maxDetectTemplateFactorNegIter,
+                            factorNeg,
+                        )
+                        if iFactorNeg > int(0.4 * self.config.maxDetectTemplateFactorNegIter):
+                            factor *= 1.25
```
kherner: Same comment here.

python/lsst/drp/tasks/assemble_coadd.py
laurenam: Yeah, it's based on a lot of testing.  With all of my experience with "adaptive" adaptations, it got to the point where I couldn't make every "magic number" configurable, but since this case is fairly simple, I can make it configurable.

kherner (approved):

PR #230, opened 2026-06-01
DM-54960: Calculate and add flags to select on non-duplicated detections in metadetect

arunkannawadi:


python/lsst/drp/tasks/metadetection_shear.py
```diff
@@ -806,6 +840,18 @@ def run(
...
+
+                self._fill_is_inner(da, sky_info.patchInfo[cell_id].inner_bbox, "is_cell_inner")
+                self._fill_is_inner(da, sky_info.patchInfo.inner_bbox, "is_patch_inner")
+                # Calculating is_tract_inner follows a different logic.
+                is_tract_inner = all(
+                    sky_info.tractInfo.inner_sky_region.contains(
+                        *sky_info.wcs.pixelToSky(Point2D(corner)).getVector()
+                    )
+                    for corner in sky_info.patchInfo.inner_bbox.getCorners()
+                )
```
TallJimbo, line 860: It looks like this logic is declaring an entire patch to be `is_tract_inner=False` if any corner of that patch is not in the tract inner region.  But that's not how tract overlaps work; they don't contain full patches.  `is_tract_inner` can only be computed per-object, not per-patch or even per-cell.

TallJimbo (commented):
I think the `is_tract_inner` calculation is wrong.

python/lsst/drp/tasks/metadetection_shear.py
TallJimbo: ~Actually, it would be fine to do this per-patch or especially per-cell _if_ it was done by testing the center of the patch or cell, because that still gets us a unique assignment.  But if per-object is okay for shear it would be better to do that for consistency with the `object` table, and if not per-cell is definitely preferable to per-patch.~

Nevermind; this ^^ is all incorrect.

arunkannawadi: There is now a correct (at least sensible) per-object `is_tract_inner` calculation that happens.

python/lsst/drp/tasks/metadetection_shear.py
```diff
@@ -947,6 +1002,18 @@ def _dictify(self, data, tract: int, patch: int):
...
+    def _fill_is_inner(
+        di: dict,
+        bbox: geom.Box2I,
+        key: str,
+    ) -> None:
+        xmin = bbox.getMinX()
+        ymin = bbox.getMinY()
+        xmax = bbox.getMaxX()
+        ymax = bbox.getMaxY()
+        di[key] = (di["x"] > xmin) & (di["x"] < xmax) & (di["y"] > ymin) & (di["y"] < ymax)
```
TallJimbo: I think this is going to abandon objects in the outermost 0.5-pixel boundary of each inner cell, since a `Box2I` has integer min and max that correspond to the _centers_ of pixels.  If you convert it to a `Box2D` this should work, but if there's any practical chance of an object landing on a half-integer position exactly, you might want to turn all the `>` into `>=`.

python/lsst/drp/tasks/metadetection_shear.py
```diff
@@ -805,6 +849,17 @@ def run(
...
                     patch=id_generator.data_id.patch.id,
                 )
+
+                self._fill_is_inner(da, sky_info.patchInfo[cell_id].inner_bbox, "is_cell_inner")
+                self._fill_is_inner(da, sky_info.patchInfo.inner_bbox, "is_patch_inner")
+                # Calculating is_tract_inner follows a different logic with
+                # regions instead of boxes.
+                da["is_tract_inner"] = sky_info.tractInfo.inner_sky_region.contains(
+                    da["ra"] * geom.degrees,
+                    da["dec"] * geom.degrees,
```
TallJimbo, line 859: These are degrees because that's something metadetect is doing intentionally?  Usually coordinates coming out of our code would be radians.

python/lsst/drp/tasks/metadetection_shear.py
arunkannawadi: Great catch. I think I'll add a +/-0.5 offset by hand and include the equality.

python/lsst/drp/tasks/metadetection_shear.py
arunkannawadi: Yes. This is in metadetect.
https://github.com/lsst/metadetect/blob/c2b1e95ebc8978f638ded7d50bc6ded1d574546e/metadetect/lsst/measure.py#L856

python/lsst/drp/tasks/metadetection_shear.py
```diff
@@ -947,6 +1002,18 @@ def _dictify(self, data, tract: int, patch: int):
...
+    @staticmethod
+    def _fill_is_inner(
+        di: dict,
+        bbox: geom.Box2I,
+        key: str,
+    ) -> None:
+        xmin = bbox.getMinX() - 0.5
+        ymin = bbox.getMinY() - 0.5
+        xmax = bbox.getMaxX() + 0.5
+        ymax = bbox.getMaxY() + 0.5
```
TallJimbo: I think conversion to `Box2D` is better than doing this by hand - note that it's still correct even if a `Box2D` is actually what's passed in, which MyPy will not flag since afw has no annotations.

But this is at least correct.

TallJimbo (approved):

python/lsst/drp/tasks/metadetection_shear.py
arunkannawadi: So, convert to `Box2D` and then dilate by 0.5?

TallJimbo: No, just convert to `Box2D` and call `getMinX()` etc. on that: the min and max of a `Box2D` are already dilated by 0.5 relative to the equivalent `Box2I`, so they cover the entirety of the integer pixels in the `Box2I`.

PR #235, opened 2026-06-25
DM-55233: Add test to validate the standardized prompt_source catalog

isullivan:
The original catalog is created by single_frame_detect_and_measure. The transform task that standardizes the catalog is in `pipe_tasks`, but the unit test needs to live here due to dependency ordering.


jrmullaney (approved):
LGTM

